PR #14, opened 2018-10-19
NN-1200 stateless with jwt

petergphillips:


petergphillips:
All now working, need to write some tests for the changes

PR #21, opened 2018-11-09
NN-1200: Copy header from licences

petergphillips:


src/main/java/uk/gov/justice/digital/hmpps/oauth2server/config/AuthenticationManagerConfiguration.java
```diff
@@ -63,12 +67,15 @@ protected void configure(final HttpSecurity http) throws Exception {
             .and().csrf().disable()
 
             .authorizeRequests()
+                .antMatchers(HttpMethod.GET, "/login").permitAll()
```
petergphillips: when you have the default failure handler spring adds in magic so that `login?error` will work.  We have to reproduce

PR #23, opened 2018-11-14
Add Oracle Specific DB Auth provider

Mjwillis:


src/main/java/uk/gov/justice/digital/hmpps/oauth2server/security/OracleAuthenticationProvider.java
```diff
@@ -0,0 +1,89 @@
...
+@Slf4j
+@Profile("oracle-auth")
+public class OracleAuthenticationProvider extends DaoAuthenticationProvider {
+
+    private final JdbcTemplate jdbcTemplate;
+    private static final String GET_HASH_DETAIL = "SELECT spare4 FROM SYS.user$ WHERE name = ?";
+    private static final String HASH = "{ ? = call SYS.DBMS_CRYPTO.hash(UTL_RAW.cast_to_raw (?) || HEXTORAW (?), DBMS_CRYPTO.hash_sh1) }";
+
+    @Autowired
+    public OracleAuthenticationProvider(DataSource dataSource, UserDetailsService userDetailsService) {
```
petergphillips: Intellij moans about this line - we have multiple databases so it thinks we should be qualifying this reference

Mjwillis: Ok will take a look at these issues on train home!

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/security/OracleAuthenticationProvider.java
```diff
@@ -38,31 +45,43 @@ public OracleAuthenticationProvider(DataSource dataSource, UserDetailsService us
...
-            logger.debug("Authentication failed: no credentials provided");
-
-            throw new BadCredentialsException(messages.getMessage(
-                    "AbstractUserDetailsAuthenticationProvider.badCredentials",
-                    "Bad credentials"));
+    protected void additionalAuthenticationChecks(final UserDetails userDetails, final UsernamePasswordAuthenticationToken authentication) throws AuthenticationException {
+        final var username = authentication.getName().toUpperCase();
+        final var password = authentication.getCredentials().toString();
+
+        if (StringUtils.isBlank(username) || StringUtils.isBlank(password)) {
```
petergphillips: I think this has got to go in the `authenticate` method instead otherwise it will fail on the user lookup instead and not give the correct message

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/security/UserDetailsServiceImpl.java
```diff
@@ -54,26 +54,38 @@ public UserDetails loadUserByUsername(final String username) throws UsernameNotF
...
+                userDetails.setAccountNonLocked(false);
+                break;
+            case LOCKED:
+                userDetails.setAccountNonLocked(false);
                 break;
             case EXPIRED:
                 userDetails.setAccountNonExpired(false);
                 break;
+            case EXPIRED_LOCKED_TIMED:
+                userDetails.setAccountNonLocked(false);
```
petergphillips: these would read better combined into single block with the four cases (actually 6) above i.e.
```java
            case EXPIRED_LOCKED_TIMED:
            case EXPIRED_GRACE_LOCKED_TIMED:
            case EXPIRED_LOCKED:
            case EXPIRED_GRACE_LOCKED:
                userDetails.setAccountNonLocked(false);
                break;
```

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/security/UserDetailsServiceImpl.java
```diff
@@ -54,26 +54,38 @@ public UserDetails loadUserByUsername(final String username) throws UsernameNotF
...
-            case OPEN:
-                userDetails.setEnabled(true);
+            case LOCKED_TIMED:
+                userDetails.setAccountNonLocked(false);
+                break;
+            case LOCKED:
+                userDetails.setAccountNonLocked(false);
                 break;
             case EXPIRED:
                 userDetails.setAccountNonExpired(false);
```
petergphillips: so for expired I think you've got to set it enabled too otherwise `AbstractUserDetailsAuthenticationProvider` will treat the account as disabled instead of expired

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/security/OracleAuthenticationProvider.java
```diff
@@ -0,0 +1,108 @@
...
+            String encodedPassword = encode(password, salt);
+
+            if (encodedPassword.equals(hash)) {
+                // reset the retry count
+                jdbcTemplate.update(RESET_LCOUNT, username);  // reset retry count
+                SecurityContextHolder.getContext().setAuthentication(authentication);
+
+            } else {
+                // check the number of retries
+                Long retryCount = jdbcTemplate.queryForObject(GET_LCOUNT, new Object[] { username }, Long.class);
```
petergphillips: is it worth combining this with the get hash statement since they are both selecting the same row from the database table?

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/security/OracleAuthenticationProvider.java
```diff
@@ -45,47 +48,49 @@ public OracleAuthenticationProvider(DataSource dataSource, UserDetailsService us
...
+                jdbcTemplate.update(UPDATE_LCOUNT, userData.getRetryCount() + 1, username);
+
                 // check the number of retries
-                Long retryCount = jdbcTemplate.queryForObject(GET_LCOUNT, new Object[] { username }, Long.class);
-                if (retryCount != null && retryCount >= 3) {
-                    // Lock the account
-                    jdbcTemplate.update(UPDATE_STATUS, AccountStatus.LOCKED.getCode(), username);
+                if (userData.getRetryCount()+1 > 2) {
+                    // Throw locked exception
+                    AccountStatus lockStatus = userData.getStatus().isGracePeriod() ? AccountStatus.EXPIRED_GRACE_LOCKED : AccountStatus.LOCKED;
```
petergphillips: Should be `LOCKED(TIMED)` I think - see http://www.dba-oracle.com/t_sysman_locked_timed_account.htm.  `LOCKED` is the manual locking of the user instead.

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/security/UserDetailsServiceImpl.java
```diff
@@ -52,28 +52,28 @@ public UserDetails loadUserByUsername(final String username) throws UsernameNotF
...
 
         AccountDetail accountDetail = userByUsername.getAccountDetail();
-        userDetails.setAccountNonExpired(!accountDetail.isExpired());
-        userDetails.setAccountNonLocked(!accountDetail.isLocked());
+        userDetails.setAccountNonExpired(true);
+        userDetails.setAccountNonLocked(true);
+        userDetails.setEnabled(false);
         userDetails.setCredentialsNonExpired(true);
 
         AccountStatus status = accountDetail.getStatus();
```
petergphillips: now you've added the new booleans to `AccountStatus` you should be able to remove this switch statement and just have:
```java
        userDetails.setAccountNonExpired(!accountDetail.getStatus().isExpired());
        userDetails.setAccountNonLocked(!accountDetail.getStatus().isLocked());
        userDetails.setEnabled(true);
        userDetails.setCredentialsNonExpired(true);
```

Mjwillis: duh yes!

PR #24, opened 2018-11-16
Java 11 / SpringBoot 2.1 / DB flyway Data

Mjwillis:


src/main/resources/application.yml
```diff
@@ -7,6 +7,8 @@ server:
     protocol_header: x-forwarded-proto
 
 spring:
+  main:
+    allow-bean-definition-overriding: true
```
Mjwillis: Added this to get it to work, TODO: work out what libs are out of date possibly

PR #30, opened 2018-11-27
NN-1437: Increase size of redirect field to hold all notm urls

petergphillips:
Also minor improvements to app too


src/main/java/uk/gov/justice/digital/hmpps/oauth2server/security/JwtAuthenticationSuccessHandler.java
```diff
@@ -25,6 +25,7 @@ public JwtAuthenticationSuccessHandler(final JwtCookieHelper jwtCookieHelper,
         this.jwtCookieHelper = jwtCookieHelper;
         this.jwtAuthenticationHelper = jwtAuthenticationHelper;
         setRequestCache(cookieRequestCache);
+        setDefaultTargetUrl("/ui");
```
petergphillips: Default to `/ui` after login

src/main/resources/application.yml
```diff
@@ -40,7 +40,10 @@ auth:
         show_sql: true
 
 logging.level.uk.gov: INFO
+
 management:
   endpoints:
     web:
       base-path: /
+    health:
+      show-details: "ALWAYS"
```
petergphillips: always show health

PR #40, opened 2018-12-04
NN-1467: Use new retries table and ALTER USER

petergphillips:


src/main/java/uk/gov/justice/digital/hmpps/oauth2server/security/H2AuthenticationProvider.java
```diff
@@ -0,0 +1,74 @@
...
+import java.time.LocalDateTime;
+
+@Component
+@Slf4j
+@Profile("oracle-auth & !oracle")
+public class H2AuthenticationProvider extends AbstractAuthenticationProvider {
+
+    private static final String GET_USER_DETAIL =
+            "SELECT p.password as spare4, 0 as retry_count, v.account_status " +
+                    "FROM v_tag_dba_users v " +
```
Mjwillis: Thinking we'll remove JPA mapping to V_TAG_DBA_USER and go straight to SYS.USER$ or DBA_USERS table . This is a syscon view which I don't think we really need to use.

PR #42, opened 2018-12-04
Fix retry table to be SQL Server Happy

Mjwillis:


Mjwillis:
Tested on local sqlserver, currently application won't build on CI, builds locally fine

PR #44, opened 2018-12-05
NN-1467: Add in UserRetries JPA entity

petergphillips:


src/main/java/uk/gov/justice/digital/hmpps/oauth2server/security/UserRetriesService.java
```diff
@@ -0,0 +1,43 @@
...
+        final var retriesOptional = repository.findById(username);
+        final UserRetries userRetries;
+        if (retriesOptional.isEmpty()) {
+            // no row exists, so create new from other table
+            userRetries = new UserRetries(username, existingRetryCount + 1);
+        } else {
+            userRetries = retriesOptional.get();
+            userRetries.incrementRetryCount();
+        }
+        repository.save(userRetries);
```
Mjwillis: Only need save on new object

PR #46, opened 2018-12-10
NN-1491: Add analytics events for authenticate and password flows

petergphillips:
Add in analytics.

Also remove Api Authentication Provider as not used any more, replaced by the new auth provider and no point adding analytics for something that isn't used.


src/main/java/uk/gov/justice/digital/hmpps/oauth2server/security/OracleAuthenticationProvider.java
```diff
@@ -31,14 +33,16 @@
...
+                                        final UserRetriesService userRetriesService,
+                                        final TelemetryClient telemetryClient) {
+        super(userDetailsService, userRetriesService, telemetryClient);
         jdbcTemplate = new JdbcTemplate(dataSource);
     }
 
     @Override
     protected UserData getUserData(final String username) {
-        return jdbcTemplate.queryForObject(GET_USER_DETAIL, new Object[]{username}, new BeanPropertyRowMapper<>(UserData.class));
+        final var results = jdbcTemplate.query(GET_USER_DETAIL, new Object[]{username}, new BeanPropertyRowMapper<>(UserData.class));
```
petergphillips: discovered that incorrect username was actually throwing an exception as `queryForObject` expects a single result

PR #54, opened 2018-12-18
NN-1511: Add in sign in details help to auth too

petergphillips:


src/main/resources/templates/login.html
```diff
@@ -26,23 +26,43 @@ <h2 class="heading-medium error-summary-heading center noMargin">
...
+                        <input class="form-control" id="username" name="username" type="text" autofocus="autofocus"
+                               data-element-id="username">
+                    </div>
+                    <div class="form-group">
+                        <label class="form-label" for="password">Password</label>
+                        <input class="form-control" id="password" name="password" type="password"
+                               data-element-id="password">
+                    </div>
                 </div>
+                <details class="govuk-details form-group">
```
jsrobertson: Is the CSS already there for the govuk-details selectors?

petergphillips: well the `details` selector was already there as we're using the css from licences, so all good :grin:

PR #58, opened 2018-12-19
NN-1528 - Allow whitelisting of /ui/** context to prevent access to m…

Mjwillis:
…anagement screens in prod/preprod

https://dsdmoj.atlassian.net/browse/NN-1528

### DIFF
--- a/src/main/java/uk/gov/justice/digital/hmpps/oauth2server/config/AuthenticationManagerConfiguration.java
+++ b/src/main/java/uk/gov/justice/digital/hmpps/oauth2server/config/AuthenticationManagerConfiguration.java
@@ -27,6 +27,7 @@ import uk.gov.justice.digital.hmpps.oauth2server.security.UserDetailsServiceImpl
 import uk.gov.justice.digital.hmpps.oauth2server.security.UserStateAuthenticationFailureHandler;
 
 import java.util.Set;
+import java.util.stream.Collectors;
 
 @Configuration
 @Order(4)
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/justice/digital/hmpps/oauth2server/config/AuthenticationManagerConfiguration.java
```diff
@@ -103,6 +112,14 @@ protected void configure(final HttpSecurity http) throws Exception {
         // @formatter:on
     }
 
+    private String buildWhitelistRestrictions() {
+        final StringBuilder whitelistAccess = new StringBuilder();
```
petergphillips: StringJoiner makes this a bit cleaner:
```java
        final var joiner = new StringJoiner(" or ", " and (", ")").setEmptyValue("");
        whitelist.forEach(s -> joiner.add(String.format("hasIpAddress('%s')", s)));
        return "isAuthenticated()" + joiner.toString();
```
there should be a nice way to stream instead, but the best I could find was:
```java
        return "isAuthenticated()" + whitelist.stream().
                map(s -> String.format("hasIpAddress('%s')", s)).
                collect(Collectors.collectingAndThen(Collectors.joining(" or "), s -> s.isEmpty() ? "" : String.format(" and (%s)", s)));
```

Mjwillis: @petergphillips updates !

PR #64, opened 2018-12-20
NN-1534: Update to latest govuk toolkit

petergphillips:


src/main/resources/application.properties
```diff
@@ -19,4 +19,5 @@ application.notify.verify.template=4b285841-41ba-47d1-b6b0-104440c4e312
 application.smoketest.enabled=false
 application.verify-email.enabled=false
 
-application.authentication.ui.whitelist=0:0:0:0:0:0:0:1,127.0.0.1/32
\ No newline at end of file
+application.authentication.ui.whitelist=0:0:0:0:0:0:0:1,127.0.0.1/32
+spring.thymeleaf.render-hidden-markers-before-checkboxes=true
```
petergphillips: magic config so thymeleaf hidden checkboxes appear before the checkbox input field and then don't break the GDS styling

jsrobertson: Nice work

PR #68, opened 2019-01-03
NN-1542: Fix display issues in ie8

petergphillips:


src/main/resources/static/css/app.css
```diff
@@ -8,6 +8,11 @@
   border-bottom: 1px solid transparent;
 }
 
+/* needed to show navigation in header for ie8 */
+#navigation ul {
+  display: block;
```
petergphillips: think there is meant to be something complicated going on with navigation, in that there is javascript to show / hide the navigation in GDS, since the styling we are overriding starts with a .js-enabled on the body.  There is something complicated going on with javascript as https://design-system.service.gov.uk/styles/page-template/custom/index.html example works in ie8 okay and it has navigation with a `ul`.  Anyway this simple css fix sorts it out for us...

PR #70, opened 2019-01-03
NN-1488: Redesign email verification pages

petergphillips:


src/main/java/uk/gov/justice/digital/hmpps/oauth2server/security/AbstractAuthenticationProvider.java
```diff
@@ -70,7 +70,7 @@ public Authentication authenticate(final Authentication authentication) throws A
             return super.authenticate(token);
         } catch (final AuthenticationException e) {
             final var reason = e.getClass().getSimpleName();
-            log.info("Authenticate failed for user {} with reason {}", username, reason);
+            log.info("Authenticate failed for user {} with reason {}", username, reason, e);
```
petergphillips: just so that we can get the exception in the logs for unknown errors

src/main/resources/static/css/app.css
```diff
@@ -22,3 +22,9 @@
 #loginForm #username {
   text-transform: uppercase;
 }
+
+#cancel {
```
petergphillips: GDS doesn't appear to have a cancel button

PR #75, opened 2019-01-08
NN-1547: First stab at show hide password

petergphillips:


src/main/resources/static/css/app.css
```diff
@@ -26,5 +26,30 @@
 #cancel {
   background-color: #dee0e2;
   color: #0b0c0c;
-  margin-left: 15px;
+}
+
+@media (min-width: 40.0625em) {
+  #cancel {
+    margin-left: 15px;
```
petergphillips: realised that we only want margin when buttons are not 100%

src/main/resources/static/css/app.css
```diff
@@ -26,5 +26,30 @@
...
+  }
+}
+
+.show-password-parent {
+  display: flex;
+}
+
+.show-password {
+  display: none;
```
petergphillips: default to not displaying so that won't be shown in ie8

PR #79, opened 2019-01-08
NN-1558: Header improvements, incl. working with IE8

petergphillips:


src/main/resources/templates/changePassword.html
```diff
@@ -18,13 +18,6 @@
         showButton.innerHTML = 'show';
       }
     }
-
-    window.addEventListener('load', function() {
```
petergphillips: since now have an ie8 specific stylesheet can set the display of the show / hide to be shown in the normal stylesheet and not have to do it with javascript

src/main/resources/static/css/app-ie8.css
```diff
@@ -0,0 +1,61 @@
+#header-container {
+  border-bottom-color: white;
```
petergphillips: yes, we are veering away from govuk styles with this change

jsrobertson: Will ask Eddie why we're not doing this.

PR #81, opened 2019-01-10
NN-1561: Wording improvements to login page and change password page

petergphillips:


src/main/resources/application.properties
```diff
@@ -12,8 +12,7 @@ logging.uris.exclude.regex=/auth/health|/auth/info|/auth/swagger.*|/auth/webjars
 application.caseload.id=NWEB
 
 application.insights.ikey=
-
-application.expired-reset.enabled=false
+application.expired-reset.enabled=true
```
petergphillips: this is now in production, so default for us should be the same

src/main/resources/messages.properties
```diff
@@ -0,0 +1,20 @@
+login.missinguser=Enter your username
```
petergphillips: since messages now appear at top of page and then for relevant field, make sense to put in separate file

PR #82, opened 2019-01-11
NN-1565: Header improvements state two

petergphillips:


src/main/resources/static/css/app-ie8.css
```diff
@@ -13,7 +13,7 @@
 }
 
 /* needed to show navigation in header for ie8 */
-#navigation ul {
+#navigation {
```
petergphillips: navigation id moved down one level to the `ul` to match gov header template

src/main/resources/static/css/app.css
```diff
@@ -22,10 +22,6 @@
   border-bottom: 1px solid transparent;
 }
 
-#navigation ul {
```
petergphillips: only needed for ie8 anyway

src/main/resources/templates/fragments/header.html
```diff
@@ -8,30 +8,29 @@
...
-
-              <image th:src="@{/images/govuk-logotype-crown.png}"
-                     class="govuk-header__logotype-crown-fallback-image"></image>
-            </svg>
+            <img class="govuk-header__logotype-crown-fallback-image" th:src="@{/images/govuk-logotype-crest.png}"
+                 alt="crest" width="42" height="35">
```
petergphillips: our crest svg was 170k so decided to just go with the png under all cases

src/main/resources/templates/fragments/header.html
```diff
@@ -8,30 +8,29 @@
...
             </span>
           </span>
         </a>
       </div>
 
-      <div class="govuk-header__content" id="header-content">
-        <nav id="navigation" sec:authorize="isAuthenticated()">
-          <ul class="govuk-header__navigation " aria-label="Top Level Navigation">
+      <div class="govuk-header__content" id="header-content" sec:authorize="isAuthenticated()">
+        <button type="button" role="button" class="govuk-header__menu-button js-header-toggle"
```
petergphillips: adding in the button etc from https://design-system.service.gov.uk/components/header/ means that we automatically get a collapsible menu option in mobile

PR #83, opened 2019-01-11
NN-1565: Add in user name to header too

petergphillips:


build.gradle
```diff
@@ -100,6 +100,7 @@ dependencies {
     compile 'org.flywaydb:flyway-core:5.2.0'
     compile 'com.zaxxer:HikariCP:2.7.6'
     compile("org.apache.commons:commons-lang3")
+    compile("org.apache.commons:commons-text:1.6")
```
petergphillips: needed for `WordUtils` as has been moved from lang3

petergphillips: so that JOE BLOGGS becomes Joe Bloggs for the header

PR #87, opened 2019-01-14
NN-1563: Add google tag manager to pages

petergphillips:


src/main/resources/templates/fragments/layout.html
```diff
@@ -13,6 +13,22 @@
 
   <meta name="theme-color" content="#0b0c0c"/>
 
+  <!-- Google Tag Manager -->
+  <script th:if="${@environment.getProperty('application.google.tag.id')} != null">(function(w, d, s, l, i) {
```
petergphillips: scripts will only be injected if application property declared

PR #89, opened 2019-01-15
NN-1203: Refactor tokens into separate table

petergphillips:
The problem with being in the USER_EMAIL table was that we had to have an email address in the table in order to create a token.
I'm planning on creating a token for the expired password scenario, but they might not have created an email address, or they might already have a token and the existing design doesn't allow for multiple tokens for the user.


src/main/resources/db/auth/V1_6__user_token.sql
```diff
@@ -0,0 +1,9 @@
+create table user_token
+(
+  token           VARCHAR(240) NOT NULL,
+  token_type      VARCHAR(10)  NOT NULL,
+  username        VARCHAR(30)  NOT NULL,
+  create_datetime DATETIME2    NOT NULL DEFAULT CURRENT_TIMESTAMP,
+  token_expiry    DATETIME2,
+  PRIMARY KEY (token)
+);
```
petergphillips: I haven't removed the old columns from this commit otherwise when we are deploying to production the server that is upgraded last won't work as columns that it was using will no longer exist

PR #98, opened 2019-01-18
NN-1584: Record locked status so can use for reset

petergphillips:


src/main/resources/db/auth/V1_7__user_email_locked.sql
```diff
@@ -0,0 +1,3 @@
+alter table user_email
+  add locked BIT NOT NULL DEFAULT 0;
```
Mjwillis: Does bit work for sql server?

petergphillips: Certainly does - `verified` is already a `BIT` type.  Translates into boolean.

PR #99, opened 2019-01-18
NN-1589: Bump lockout to 10 and make configurable

petergphillips:


src/main/resources/application-dev.yml
```diff
@@ -23,6 +23,9 @@ application:
   reset-password:
     enabled: true
 
+  authentication:
+    lockout-count: 3
```
petergphillips: so that integration tests continue to lockout on 3 attempts

src/main/resources/templates/changePassword.html
```diff
@@ -33,8 +33,12 @@ <h2 class="govuk-error-summary__title" id="error-summary">
     <div class="govuk-error-summary__body" id="error-detail">
       <ul class="govuk-list govuk-error-summary__list">
 
-        <li th:unless="${#arrays.isEmpty(param.errorcurrent)}" th:each="error: ${param.errorcurrent}">
-          <a th:text="#{${'changepassword.' + error}}" href="#password">Current password message</a>
+        <li th:unless="${#arrays.isEmpty(param.errorcurrent)}" th:each="error: ${param.errorcurrent}"
```
petergphillips: this is a bit naff, but will disappear as will soon not need current password for the expired password scenario

src/main/resources/templates/login.html
```diff
@@ -34,6 +34,9 @@ <h2 class="govuk-error-summary__title" id="error-summary">
       <ul class="govuk-list govuk-error-summary__list">
         <li th:each="error: ${param.error}" th:switch="${error}">
           <a th:case="'missingpass'" href="#password" th:text="#{login.missingpass}">Missing password</a>
+          <a th:case="'invalid'" href="#username"
+             th:text="#{login.invalid(${@environment.getProperty('application.authentication.lockout-count')})}">
```
petergphillips: thought it was cleaner to directly inject application property this way than adding to the model in the controller

PR #101, opened 2019-01-18
NN-1203: More work on forgotten password

petergphillips:


src/main/java/uk/gov/justice/digital/hmpps/oauth2server/nomis/model/AccountStatus.java
```diff
@@ -49,11 +54,15 @@ public boolean isGracePeriod() {
         return gracePeriod;
     }
 
-    public static AccountStatus get(int code) {
+    public boolean isUserLocked() {
+        return userLocked;
```
petergphillips: we need to know if it is locked by the user getting the password wrong in c-nomis

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/verify/ResetPasswordService.java
```diff
@@ -0,0 +1,106 @@
...
+        if (userOptional.isEmpty()) {
+            emailTemplate = resetUnavailableTemplateId;
+            parameters = Map.of("firstName", username);
+        } else {
+            final var staffUserAccount = userOptional.get();
+            final var status = staffUserAccount.getAccountDetail().getStatus();
+            // only allow reset for active accounts that aren't locked
+            // or are locked by getting password incorrect (in either c-nomis or auth)
+            final var firstName = staffUserAccount.getStaff().getFirstName();
+            if (staffUserAccount.getStaff().isActive() && (!status.isLocked() || status.isUserLocked() || userEmail.isLocked())) {
```
petergphillips: bit complicated here - need to cope with users that have locked themselves out of c-nomis _or_ users that have locked themselves out using auth as unfortunately the status will be different (auth can't set the account status to user locked so has to have own flag instead)

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/verify/ResetPasswordService.java
```diff
@@ -0,0 +1,106 @@
...
+                // delete any existing token
+                userTokenOptional.ifPresent(userTokenRepository::delete);
+
+                final var userToken = new UserToken(TokenType.RESET, userEmail);
+                userTokenRepository.save(userToken);
+
+                final var resetLink = String.format("%s/%s", url, userToken.getToken());
+                parameters = Map.of("firstName", firstName, "resetLink", resetLink);
+            } else {
+                emailTemplate = resetUnavailableTemplateId;
```
petergphillips: we send them an email if we have their email address, but can't reset their password.  Better than divulging information about their account on the website.

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/verify/ResetPasswordService.java
Mjwillis: yup bit complex - not much we can do at moment thou,  maybe this logic could be in one place for future commit.

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/verify/ResetPasswordService.java
```diff
@@ -0,0 +1,106 @@
...
+        final var email = userEmail.getEmail();
+        try {
+            log.info("Sending reset password to notify for user {}", username);
+            notificationClient.sendEmail(emailTemplate, email, parameters, null);
+            telemetryClient.trackEvent("ResetPasswordRequestSuccess", Map.of("username", username), null);
+        } catch (final NotificationClientException e) {
+            final var reason = (e.getCause() != null ? e.getCause() : e).getClass().getSimpleName();
+            log.warn("Failed to send reset password to notify for user {}", username, e);
+            telemetryClient.trackEvent("ResetPasswordRequestFailure", Map.of("username", username, "reason", reason), null);
+            if (e.getHttpResult() >= 500) {
```
Mjwillis: maybe we add a retry camel route for this in future.

PR #104, opened 2019-01-21
NN-1203: Re-check locked status before allowing reset password

petergphillips:


src/main/java/uk/gov/justice/digital/hmpps/oauth2server/verify/ResetPasswordService.java
```diff
@@ -132,6 +139,14 @@ public ResetPasswordService(final UserEmailRepository userEmailRepository,
     public void resetPassword(final String token, final String newPassword) {
         final var userToken = userTokenRepository.findById(token).orElseThrow();
         final var userEmail = userToken.getUserEmail();
+        final var userOptional = userService.getUserByUsername(userEmail.getUsername());
+
+        // before we reset, ensure user allowed to still reset password
+        if (userOptional.isEmpty() || !passwordAllowedToBeReset(userEmail, userOptional.get())) {
```
petergphillips: we checked they were allowed to reset when we sent email, but 23.59 hours could have passed since then, so best to recheck before doing the reset

src/main/resources/messages.properties
```diff
@@ -1,7 +1,7 @@
 login.missinguser=Enter your username
 login.missingpass=Enter your password
 login.expired=Your password has expired
-login.locked=Your account is locked. To unlock it, reset your password on Classic NOMIS.
+login.locked=Your account is locked. If you have verified your email address then you can use 'I have forgotten my password' below.
```
petergphillips: we can now let them know they can reset their password using the new feature :grin:

PR #119, opened 2019-01-28
NN-1545: Add in oracle password encoder

petergphillips:


src/main/java/uk/gov/justice/digital/hmpps/oauth2server/config/PasswordConfig.java
```diff
@@ -3,11 +3,20 @@
...
+
+import java.util.Map;
 
 @Configuration
 public class PasswordConfig {
     @Bean
-    public BCryptPasswordEncoder passwordEncoder() {
-        return new BCryptPasswordEncoder();
+    public PasswordEncoder passwordEncoder() {
+        final var bCryptPasswordEncoder = new BCryptPasswordEncoder();
```
petergphillips: Unfortunately can't use default `PasswordEncoderFactories` out of the box since we want to add an oracle password encoder too

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/security/PasswordEncoderUtils.java
```diff
@@ -0,0 +1,59 @@
...
+package uk.gov.justice.digital.hmpps.oauth2server.security;
+
+import org.springframework.security.crypto.codec.Utf8;
+
+/**
+ * Utility for constant time comparison to prevent against timing attacks.
+ *
+ * @author Rob Winch
+ */
+class PasswordEncoderUtils {
```
petergphillips: Annoyingly Spring make this package private so had to copy instead

PR #120, opened 2019-01-28
NN-1545: Fix password encoder for authentication

petergphillips:


src/main/java/uk/gov/justice/digital/hmpps/oauth2server/security/AbstractAuthenticationProvider.java
```diff
@@ -24,11 +25,13 @@
     public AbstractAuthenticationProvider(final UserDetailsService userDetailsService,
                                           final UserRetriesService userRetriesService,
                                           final TelemetryClient telemetryClient,
+                                          final PasswordEncoder passwordEncoder,
                                           final int accountLockoutCount) {
         this.userRetriesService = userRetriesService;
         this.telemetryClient = telemetryClient;
         this.accountLockoutCount = accountLockoutCount;
         setUserDetailsService(userDetailsService);
+        setPasswordEncoder(passwordEncoder);
```
petergphillips: think I removed this, thinking it wasn't needed, but the superclass (`DaoAuthenticationProvider`) explicitly creates its own password encoder instead, which we then need to override

PR #124, opened 2019-01-29
NN-1545: Stop password123456 as a password, erase credentials from us…

petergphillips:
…er details if using external id and rename abstract class that is no longer abstract


src/main/java/uk/gov/justice/digital/hmpps/oauth2server/security/ExternalIdAuthenticationHelper.java
```diff
@@ -17,52 +18,57 @@
...
 
-            StaffUserAccount userDetail = userService.getUserByUsername(username).orElseThrow(() -> new OAuth2AccessDeniedException("No user found matching username."));
+            final var userDetail = userService.getUserByUsername(username).orElseThrow(() -> new OAuth2AccessDeniedException("No user found matching username."));
             // Get full user details, with authorities, etc.
             userDetails = userDetailsService.loadUserByUsername(userDetail.getUsername());
+        } else {
+            userDetails = null;
         }
 
+        if (userDetails instanceof CredentialsContainer) {
```
petergphillips: since we are returning user details here should really erase credentials

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/security/LockingAuthenticationProvider.java
```diff
@@ -19,16 +19,16 @@
 
 @Slf4j
 @Component
-public class AbstractAuthenticationProvider extends DaoAuthenticationProvider {
+public class LockingAuthenticationProvider extends DaoAuthenticationProvider {
```
petergphillips: no longer abstract!

PR #125, opened 2019-01-29
NN-1545: Initial stab at auth login

petergphillips:


src/main/resources/db/auth/V1_8__user_email_fields.sql
```diff
@@ -0,0 +1,26 @@
+alter table user_email
+  add password VARCHAR(100);
+alter table user_email
+  add enabled BIT NOT NULL DEFAULT 0;
+alter table user_email
+  add master BIT NOT NULL DEFAULT 0;
```
petergphillips: use `master` to indicate that we are the master of the data, for users that exist in both nomis and auth

PR #130, opened 2019-01-31
NN-1622: Allow auth only users to change password and verify email

petergphillips:


src/main/java/uk/gov/justice/digital/hmpps/oauth2server/nomis/model/StaffUserAccount.java
```diff
@@ -54,4 +63,43 @@
...
+
+    @Override
+    public String getFirstName() {
+        return getStaff().getFirstName();
+    }
+
+
+    @Override
+    public Collection<? extends GrantedAuthority> getAuthorities() {
+        return filterRolesByCaseload("NWEB").stream()
```
petergphillips: note that I've hard coded `NWEB` here, don't think it will change.  Otherwise I would have to inject a spring variable into a model, which seems weird

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/security/UserDetailsServiceImpl.java
```diff
@@ -1,101 +1,27 @@
...
-                accountNonLocked = false;
-                break;
-            default:
-                enabled = false;
-                credentialsNonExpired = true;
-                accountNonLocked = true;
-        }
-
-        // when running using oracle for nomis datasource then passwords need prefixing when retrieved
-        final var password = (oracleForNomis ? "{oracle}" : "") + staffUserAccount.getPassword();
```
petergphillips: we don't need this any more since the new default password encoder is the oracle one

PR #133, opened 2019-02-01
NN-1622: Expire auth only users and test change password

petergphillips:


src/main/resources/db/auth/V1_9__user_email_fields_2.sql
```diff
@@ -0,0 +1,2 @@
+alter table user_email
+  add password_expiry DATETIME NOT NULL DEFAULT CURRENT_TIMESTAMP;
```
petergphillips: I've tested against sqlserver too

src/main/resources/application.properties
```diff
@@ -26,6 +26,8 @@ application.hdc-endpoint-url=
 
 application.authentication.ui.whitelist=0:0:0:0:0:0:0:1,127.0.0.1/32,172.16.0.0/12
 application.authentication.lockout-count=10
+application.authentication.password-age=3650
```
petergphillips: default password age to 10 years until told otherwise :grin:

PR #137, opened 2019-02-07
NN-1638: Add in retrieving auth users by client credentials and passw…

petergphillips:
…ord grant

### DIFF
--- a/src/main/java/uk/gov/justice/digital/hmpps/oauth2server/security/ExternalIdAuthenticationHelper.java
+++ b/src/main/java/uk/gov/justice/digital/hmpps/oauth2server/security/ExternalIdAuthenticationHelper.java
@@ -1,9 +1,9 @@
 package uk.gov.justice.digital.hmpps.oauth2server.security;
 
 import org.apache.commons.lang3.StringUtils;
-import org.springframework.security.core.CredentialsContainer;
 import org.springframework.security.core.userdetails.UserDetails;
 import org.springframework.security.core.userdetails.UserDetailsService;
+import org.springframework.security.core.userdetails.UsernameNotFoundException;
 import org.springframework.security.oauth2.client.resource.OAuth2AccessDeniedException;
 import org.springframework.stereotype.Component;
 import uk.gov.justice.digital.hmpps.oauth2server.nomis.model.StaffUserAccount;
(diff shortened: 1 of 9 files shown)


src/main/java/uk/gov/justice/digital/hmpps/oauth2server/security/ExternalIdAuthenticationHelper.java
```diff
@@ -51,24 +49,21 @@ public UserDetails getUserDetails(final Map<String, String> requestParameters) {
...
-            final var userDetail = userService.getUserByUsername(username).orElseThrow(() -> new OAuth2AccessDeniedException("No user found matching username."));
-            // Get full user details, with authorities, etc.
-            userDetails = userDetailsService.loadUserByUsername(userDetail.getUsername());
-        } else {
-            userDetails = null;
-        }
-
-        if (userDetails instanceof CredentialsContainer) {
-            ((CredentialsContainer) userDetails).eraseCredentials();
+            try {
```
petergphillips: don't need separate call as both return same user details object these days

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/security/UserDetailsServiceImpl.java
```diff
@@ -8,20 +8,39 @@
...
         this.userService = userService;
     }
 
     @Override
     @Cacheable("loadUserByUsername")
     public UserDetails loadUserByUsername(final String username) throws UsernameNotFoundException {
-        return userService.findUser(username).orElseThrow(() -> new UsernameNotFoundException(username));
+        final var userPersonDetails = userService.findUser(username).orElseThrow(() -> new UsernameNotFoundException(username));
+
+        // ensure that any changes to user details past this point are not persisted - e.g. by calling
```
petergphillips: Open for discussion here.  I've gone down the detach object approach.  The other one would be to add in a clone that did a `toBuilder().build()` under the covers.  The clone approach seemed a bit nasty as any lazy collections that we don't need would automatically get instantiated.  This also seems nasty as we need to delve into the object to work out which persistence unit it belongs to...

src/main/resources/application.properties
```diff
@@ -24,7 +24,7 @@ application.hdc-endpoint-url=
 
 application.authentication.ui.whitelist=0:0:0:0:0:0:0:1,127.0.0.1/32,172.16.0.0/12
 application.authentication.lockout-count=10
-application.authentication.password-age=3650
+application.authentication.password-age=7300
```
petergphillips: add in better sensible default for password age

PR #138, opened 2019-02-07
NN-1638: Add in authentication source so elite2 can tell where user o…

petergphillips:
…riginated from

### DIFF
--- a/src/main/java/uk/gov/justice/digital/hmpps/oauth2server/auth/model/UserEmail.java
+++ b/src/main/java/uk/gov/justice/digital/hmpps/oauth2server/auth/model/UserEmail.java
@@ -105,4 +105,9 @@ public class UserEmail implements UserPersonDetails, CredentialsContainer {
     public boolean isAdmin() {
         return false;
     }
+
+    @Override
+    public String getAuthSource() {
+        return "auth";
+    }
 }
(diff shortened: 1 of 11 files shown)


src/main/java/uk/gov/justice/digital/hmpps/oauth2server/auth/model/UserEmail.java
```diff
@@ -105,4 +105,9 @@ public String getFirstName() {
     public boolean isAdmin() {
         return false;
     }
+
+    @Override
+    public String getAuthSource() {
+        return "auth";
```
petergphillips: for now this is a string, if it transpires that it needs testing to see where user is from then will change to enum

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/security/ExternalIdAuthenticationHelper.java
```diff
@@ -30,40 +28,42 @@ public ExternalIdAuthenticationHelper(final UserService userService, final UserD
...
     }
 
-    public UserDetails getUserDetails(final Map<String, String> requestParameters) {
+    public UserPersonDetails getUserDetails(final Map<String, String> requestParameters) {
         if (requestParameters.containsKey(REQUEST_PARAM_USER_ID_TYPE) &&
                 requestParameters.containsKey(REQUEST_PARAM_USER_ID)) {
-            // Extract values - if either are empty/null, throw auth failed exception
-            final var userIdType = requestParameters.get(REQUEST_PARAM_USER_ID_TYPE);
-            final var userId = requestParameters.get(REQUEST_PARAM_USER_ID);
+            return loadByUserIdType(requestParameters.get(REQUEST_PARAM_USER_ID_TYPE), requestParameters.get(REQUEST_PARAM_USER_ID));
```
petergphillips: tidying after discussion with @JamieDraperUK

PR #141, opened 2019-02-08
NN-1638: Fix auth source not being set and test

petergphillips:


src/test/groovy/uk/gov/justice/digital/hmpps/oauth2server/integration/specs/LoginSpecification.groovy
```diff
@@ -177,26 +214,71 @@ class LoginSpecification extends GebReportingSpec {
...
+        and: 'state is returned'
+        browser.getCurrentUrl() contains("state=$state")
+
+        and: 'auth code is returned'
+        def params = splitQuery(new URL(browser.getCurrentUrl()))
+        def authCode = params.get('code')
+        authCode != null
+
+        and: 'auth code can be redeemed for access token'
+        def response = getAccessToken(authCode)
```
petergphillips: actually test that the authentication code can be redeemed by a server to server call

PR #144, opened 2019-02-11
NN-1641: Inform user expected password length

petergphillips:


src/main/java/uk/gov/justice/digital/hmpps/oauth2server/resource/AbstractPasswordController.java
```diff
@@ -164,4 +175,12 @@ public AbstractPasswordController(final PasswordService passwordService,
         modelAndView.addObject("error", Boolean.TRUE);
         return Optional.of(modelAndView);
     }
+
+    private void addIsAdminToModel(@SuppressWarnings("OptionalUsedAsFieldOrParameterType") final Optional<UserToken> userToken,
+                                   final ModelAndView modelAndView) {
+        final boolean isAdmin = userToken.flatMap(ut -> userService.findUser(ut.getUserEmail().getUsername())).
```
petergphillips: `flatMap` here is quite cool as `findUser` returns an optional too

PR #158, opened 2019-03-04
LIC-907: Move validation to service so can be shared and minor other …

petergphillips:
…fixes

### DIFF
--- a/src/main/java/uk/gov/justice/digital/hmpps/oauth2server/resource/ResetPasswordController.java
+++ b/src/main/java/uk/gov/justice/digital/hmpps/oauth2server/resource/ResetPasswordController.java
@@ -7,7 +7,6 @@ import org.springframework.beans.factory.annotation.Value;
 import org.springframework.stereotype.Controller;
 import org.springframework.validation.annotation.Validated;
 import org.springframework.web.bind.annotation.GetMapping;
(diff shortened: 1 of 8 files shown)


src/main/java/uk/gov/justice/digital/hmpps/oauth2server/resource/ResetPasswordController.java
```diff
@@ -90,13 +90,6 @@ public ModelAndView resetPasswordRequest(@RequestParam(required = false) final S
         }
     }
 
-    @GetMapping("/reset-password-confirm/{token}")
-    @Deprecated
```
petergphillips: old code, not used after go live a month ago

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/verify/ResetPasswordService.java
```diff
@@ -131,6 +131,7 @@ public void setPassword(final String token, final String password) {
             throw new LockedException("locked");
         }
 
+        userEmail.setVerified(true);
```
petergphillips: we've allowed them to set a new password with an unverified email address, so should mark it as verified now

PR #165, opened 2019-03-06
NN-1639: Experiment with throttling requests on reset password

petergphillips:


src/main/java/uk/gov/justice/digital/hmpps/oauth2server/verify/PasswordService.java
```diff
@@ -1,29 +1,9 @@
 package uk.gov.justice.digital.hmpps.oauth2server.verify;
 
-import org.springframework.security.crypto.password.PasswordEncoder;
 import uk.gov.justice.digital.hmpps.oauth2server.auth.model.UserEmail;
-import uk.gov.justice.digital.hmpps.oauth2server.security.ReusedPasswordException;
 
-import java.time.LocalDateTime;
+public interface PasswordService {
```
petergphillips: unfortunately the jdk proxying needed for the throttling requires an interface

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/verify/ResetPasswordServiceImpl.java
```diff
@@ -0,0 +1,151 @@
...
+        this.userTokenRepository = userTokenRepository;
+        this.userService = userService;
+        this.alterUserService = alterUserService;
+        this.notificationClient = notificationClient;
+        this.resetTemplateId = resetTemplateId;
+        this.resetUnavailableTemplateId = resetUnavailableTemplateId;
+    }
+
+    @Override
+    @Throttling(limit = 6, timeUnit = TimeUnit.MINUTES)
```
petergphillips: Allow 6 requests from the same ip address every minute

PR #180, opened 2019-03-20
LIC-922: Add optional vary when creating licence user and upgrade spr…

petergphillips:
…ing to latest and greatest


src/test/groovy/uk/gov/justice/digital/hmpps/oauth2server/integration/specs/ResetPasswordSpecification.groovy
```diff
@@ -44,7 +44,7 @@ class ResetPasswordSpecification extends GebReportingSpec {
 
         when: 'The Reset Password page is displayed'
         at ResetPasswordPage
-        resetPasswordAs CA_USER
+        resetPasswordAs RESET_TEST_USER
```
petergphillips: If the reset password test ran before the landing page login test then the landing page test would fail.  Now using separate users

PR #183, opened 2019-03-28
NN-1739: Add in maintenance of user roles

petergphillips:


src/main/java/uk/gov/justice/digital/hmpps/oauth2server/resource/AuthUserController.java
```diff
@@ -51,7 +51,7 @@ public AuthUserController(final UserService userService, final CreateUserService
             @ApiResponse(code = 401, message = "Unauthorized.", response = ErrorDetail.class),
             @ApiResponse(code = 404, message = "User not found.", response = ErrorDetail.class)})
     public ResponseEntity<Object> user(@ApiParam(value = "The username of the user.", required = true) @PathVariable final String username) {
-        final var user = userService.getAuthUserByUsername(StringUtils.trim(username));
```
petergphillips: moved to service as other services need to make same calls

src/main/resources/db/dev/data/auth/V9_1__oauth_sample_data.sql
```diff
@@ -74,6 +74,6 @@ VALUES ('36025454-e42d-49a1-9124-013577a7ed20', 'AUTH_ADM', 'ROLE_OAUTH_ADMIN'),
        ('a4843bf0-9b44-451c-ba3e-cdf04ba9eb3c', 'AUTH_RO_USER', 'ROLE_GLOBAL_SEARCH'),
        ('b4843bf0-9b44-451c-ba3e-cdf04ba9eb3b', 'AUTH_RO_VARY_USER', 'ROLE_LICENCE_RO'),
        ('b4843bf0-9b44-451c-ba3e-cdf04ba9eb3c', 'AUTH_RO_VARY_USER', 'ROLE_GLOBAL_SEARCH'),
-       ('b4843bf0-9b44-451c-ba3e-cdf04ba9eb3c', 'AUTH_RO_VARY_USER', 'ROLE_LICENCE_VARY'),
+       ('b4843bf0-9b44-451c-ba3e-cdf04ba9eb3d', 'AUTH_RO_VARY_USER', 'ROLE_LICENCE_VARY'),
```
petergphillips: need different id otherwise doesn't get inserted

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/auth/model/Authority.java
```diff
@@ -25,6 +26,15 @@
     private String authority;
 
     public Authority(final String authority) {
-        this.authority = authority;
+        this.authority = addRolePrefixIfNecessary(authority);
+    }
+
+    public String getAuthorityName() {
+        // strip off ROLE_
```
JamieDraperUK: You might be able to remove some duplication and remove this comment if ROLE_ was stored in a constant and was referenced throughout. Then you could write return authority.substring(ROLE_A_GOOD_NAME);

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/resource/AuthUserRolesController.java
```diff
@@ -0,0 +1,122 @@
...
+            @ApiResponse(code = 400, message = "Validation failed.", response = ErrorDetail.class),
+            @ApiResponse(code = 401, message = "Unauthorized.", response = ErrorDetail.class),
+            @ApiResponse(code = 404, message = "User not found.", response = ErrorDetail.class),
+            @ApiResponse(code = 409, message = "Role for user already exists.", response = ErrorDetail.class),
+            @ApiResponse(code = 500, message = "Server exception e.g. failed to insert row.", response = ErrorDetail.class)})
+    public ResponseEntity<Object> addRole(
+            @ApiParam(value = "The username of the user.", required = true) @PathVariable final String username,
+            @ApiParam(value = "The role to be added to the user.", required = true) @PathVariable final String role) {
+
+        final var userOptional = userService.getAuthUserByUsername(username);
```
JamieDraperUK: Would this logic be better of in the service?

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/auth/model/Authority.java
petergphillips: done

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/resource/AuthUserRolesController.java
petergphillips: Tricky one.

The problem is that we return responses with different codes etc. depending on whether the user exists, whether the role exists etc.

One solution is to use magic to solve the problem, so the services throw exceptions of a certain type which then automatically get converted into the correct response code etc.  We haven't gone down that route here so there aren't any global exception handlers that cope with exceptions, instead the controller deals with the response manually.  It has the advantage that it is clear what is going on at all times, but it does lead to the controller having more logic

JamieDraperUK: Well, you could throw the correct exceptions from your service, then handle them right here in this controller method.

PR #187, opened 2019-04-17
NN-1834: Add field in error to error response and upgrade libraries

petergphillips:


build.gradle
```diff
@@ -1,21 +1,22 @@
...
+        classpath "com.github.ben-manes:gradle-versions-plugin:0.21.0"
     }
-
 }
 
 apply plugin: 'java'
 apply plugin: 'org.springframework.boot'
 apply plugin: 'io.spring.dependency-management'
 apply plugin: 'groovy'
+apply plugin: "com.github.ben-manes.versions"
```
petergphillips: Adds in the `dependencyUpdates` task to gradle so can tell what versions can be updated

PR #201, opened 2019-06-04
NN-1966: Make initial password creation service agnostic

petergphillips:


src/main/java/uk/gov/justice/digital/hmpps/oauth2server/resource/HomeController.java
```diff
@@ -1,52 +1,26 @@
...
-        this.nnUrl = nnUrl;
-        this.hdcUrl = hdcUrl;
     }
 
     @GetMapping("/")
     public ModelAndView home(final Authentication authentication) {
         final var services = landingService.findAllServices();
 
-        // fallback to current behaviour if no services defined
-        if (services.isEmpty()) {
```
petergphillips: bit of a tidy up - services now defined in all environments, so don't need fall back any more

PR #205, opened 2019-06-10
DT-16: Fix api to show public key

petergphillips:


src/main/java/uk/gov/justice/digital/hmpps/oauth2server/resource/api/PublicKeyController.java
```diff
@@ -37,26 +36,26 @@ public PublicKeyController(@Value("${jwt.signing.key.pair}") final String privat
...
-        return result;
+    private String[] convertNewLinesToArray(final String formattedKey) {
+        return formattedKey.split("\n");
     }
 
     private String getFormattedKey(final PublicKey pk) {
         final var builder = new StringBuilder();
         final var encodeBase64String = Base64.encodeBase64String(pk.getEncoded());
         builder.append("-----BEGIN PUBLIC KEY-----");
-        builder.append("\r\n");
```
petergphillips: public key should only use `\n`

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/resource/api/PublicKeyController.java
```diff
@@ -37,26 +36,26 @@ public PublicKeyController(@Value("${jwt.signing.key.pair}") final String privat
...
+        builder.append("\n");
         for (var i = 0; i < encodeBase64String.length(); i += 64) {
-            builder.append(encodeBase64String.substring(i, Math.min(i + 64, encodeBase64String.length())));
-            builder.append("\r\n");
+            builder.append(encodeBase64String, i, Math.min(i + 64, encodeBase64String.length()));
+            builder.append("\n");
```
petergphillips: and should have one at end of key

PR #206, opened 2019-06-10
DT-16: Correct formatted display and add integration test

petergphillips:


src/main/java/uk/gov/justice/digital/hmpps/oauth2server/resource/api/PublicKeyController.java
```diff
@@ -38,11 +38,7 @@ public PublicKeyController(@Value("${jwt.signing.key.pair}") final String privat
     @ApiResponses(value = {@ApiResponse(code = 200, message = "OK", response = Map.class)})
     public Map<String, Object> getJwtPublicKey() {
         final var formattedKey = getFormattedKey(publicKey);
-        return Map.of("formatted", convertNewLinesToArray(formattedKey), "encoded", Base64.encodeBase64String(formattedKey.getBytes()));
-    }
-
-    private String[] convertNewLinesToArray(final String formattedKey) {
-        return formattedKey.split("\n");
```
petergphillips: can't be an array as with json there is no guarantee of order of elements

petergphillips: :man_facepalming: was getting confused with objects.  In json the array order is guaranteed.  No idea why my test was failing previously when I tried it, but working now :woman_shrugging:

PR #208, opened 2019-06-19
DT-63: Add new pecs roles for add user and upgrade libs etc.

petergphillips:


.circleci/config.yml
```diff
@@ -124,3 +124,7 @@ workflows:
       - build_docker:
           requires:
             - integration_tests
+          filters:
```
petergphillips: only need docker images for auth on master

Dockerfile
```diff
@@ -1,13 +1,23 @@
...
+    adduser --uid 2000 --system appuser --gid 2000
+
 WORKDIR /app
 
 COPY build/libs/oauth2server*.jar /app/app.jar
 COPY run.sh /app
 
-ENV TZ=Europe/London
-RUN ln -snf "/usr/share/zoneinfo/$TZ" /etc/localtime && echo "$TZ" > /etc/timezone
+RUN chown -R appuser:appgroup /app
```
petergphillips: updated to latest config so run as normal user etc

build.gradle
```diff
@@ -1,46 +1,38 @@
...
+import org.springframework.boot.gradle.plugin.SpringBootPlugin
+
+import java.time.Instant
+import java.time.LocalDate
+import java.time.format.DateTimeFormatter
+
+plugins {
+    id 'java'
+    id 'groovy'
+    id 'org.springframework.boot' version '2.1.6.RELEASE'
```
petergphillips: latest spring etc

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/maintain/AuthUserRoleService.java
```diff
@@ -16,7 +16,12 @@
 @Service
 @Slf4j
 public class AuthUserRoleService {
-    static final Map<String, String> ALLOWED_AUTH_USER_ROLES = Map.of("ROLE_LICENCE_VARY", "Licence Variation", "ROLE_LICENCE_RO", "Licence Responsible Officer", "ROLE_GLOBAL_SEARCH", "Global Search");
+    static final Map<String, String> ALLOWED_AUTH_USER_ROLES = Map.of(
+            "ROLE_LICENCE_VARY", "Licence Variation",
+            "ROLE_LICENCE_RO", "Licence Responsible Officer",
+            "ROLE_GLOBAL_SEARCH", "Global Search",
+            "ROLE_PECS_POLICE", "PECS Police",
```
petergphillips: new pecs roles

Dockerfile
andrewrlee: think you can do the chown as part of copy? https://stackoverflow.com/a/44766666

build.gradle
```diff
@@ -62,84 +54,84 @@ springBoot {
...
+        mavenBom(SpringBootPlugin.BOM_COORDINATES)
+    }
+}
+
 dependencies {
     annotationProcessor 'org.springframework.boot:spring-boot-configuration-processor'
     annotationProcessor('org.projectlombok:lombok:1.18.8')
 
-    compile files('lib/ojdbc8-18.3.jar')
+    implementation files('lib/ojdbc8-18.3.jar')
```
andrewrlee: can use `testAnnotationProcessor` to enable lombok in test...

PR #211, opened 2019-06-26
DT-8 Return plain/text from ping controller.

andrewrlee:
Superagent failed to parse the raw value 'pong' as json

### DIFF
--- a/src/main/java/uk/gov/justice/digital/hmpps/oauth2server/resource/api/PingEndpoint.java
+++ b/src/main/java/uk/gov/justice/digital/hmpps/oauth2server/resource/api/PingEndpoint.java
@@ -3,12 +3,16 @@ package uk.gov.justice.digital.hmpps.oauth2server.resource.api;
 import org.springframework.boot.actuate.endpoint.annotation.ReadOperation;
 import org.springframework.boot.actuate.endpoint.web.annotation.WebEndpoint;
 import org.springframework.context.annotation.Configuration;
+import org.springframework.http.MediaType;
+import org.springframework.http.ResponseEntity;
+
+import static org.springframework.http.MediaType.*;
 
 @Configuration
 @WebEndpoint(id = "ping")
 public class PingEndpoint {
     @ReadOperation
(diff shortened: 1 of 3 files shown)


src/test/java/uk/gov/justice/digital/hmpps/oauth2server/resource/api/PingEndpointTest.java
```diff
@@ -1,12 +1,15 @@
...
 
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.springframework.http.MediaType.TEXT_PLAIN;
 
 public class PingEndpointTest {
     @Test
     public void ping() {
-        assertThat(new PingEndpoint().ping()).isEqualTo("pong");
+        assertThat(new PingEndpoint().ping()).isEqualTo(
+                ResponseEntity.ok().contentType(TEXT_PLAIN).body("pong"));
```
petergphillips: not a great fan of the test duplicating the production code, doesn't really test anything.  That said, it was doing it before (albeit for a string) :man_shrugging:

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/resource/api/PingEndpoint.java
```diff
@@ -3,12 +3,16 @@
...
 import org.springframework.boot.actuate.endpoint.web.annotation.WebEndpoint;
 import org.springframework.context.annotation.Configuration;
+import org.springframework.http.MediaType;
+import org.springframework.http.ResponseEntity;
+
+import static org.springframework.http.MediaType.*;
 
 @Configuration
 @WebEndpoint(id = "ping")
 public class PingEndpoint {
```
JamieDraperUK: Could you use a spring annotation to solve this problem? Something like @Produces(value = "text/plain")

petergphillips: :trophy:

PR #218, opened 2019-07-10
DT-74: Initial stab at group maintenance

petergphillips:


src/main/java/uk/gov/justice/digital/hmpps/oauth2server/maintain/AuthUserGroupService.java
```diff
@@ -0,0 +1,106 @@
...
+        }
+
+        // TODO: Validate that the group is allowed to be added to the user:
+        // 1. If super user then can add anyone to anything
+        // 2. If group admin then needs to be one of their groups and user can't be a member of a different group
+
+        log.info("Adding group {} to user {}", groupFormatted, username);
+        userEmail.getGroups().add(group);
+        telemetryClient.trackEvent("AuthUserGroupAddSuccess", Map.of("username", username, "group", groupFormatted, "admin", modifier), null);
+        userEmailRepository.save(userEmail);
```
andrewrlee: are these present just for tests? would the transaction automatically propagate the change to the db?

petergphillips: good point.  Initially they were there because it wasn't working without them, but I then found out that the real reason was I was using `@Transactional` with the wrong transaction manager :woman_facepalming: .  I'll remove :grin:

src/test/java/uk/gov/justice/digital/hmpps/oauth2server/maintain/AuthUserGroupServiceTest.java
```diff
@@ -0,0 +1,113 @@
...
+        assertThat(groups).isNotPresent();
+    }
+
+    @Test
+    public void getAuthUserGroups_success() {
+        final var user = new UserEmail("user");
+        user.setGroups(new HashSet<>(List.of(new Group("JOE", "desc"), new Group("LICENCE_VARY", "desc2"))));
+        when(userEmailRepository.findByUsernameAndMasterIsTrue(anyString())).thenReturn(Optional.of(user));
+        final var groups = service.getAuthUserGroups(" BOB ");
+        //noinspection OptionalGetWithoutIsPresent
```
andrewrlee: meh...`getOrThrow` would remove warning

petergphillips: oops sorry forgot about that

petergphillips: will fix on next PR

PR #219, opened 2019-07-10
DT-74: Pass group too on create and get assignable groups

petergphillips:


src/main/java/uk/gov/justice/digital/hmpps/oauth2server/resource/api/AuthUserController.java
```diff
@@ -200,12 +217,8 @@ private String createInitialPasswordUrl(@ApiIgnore final HttpServletRequest requ
         private String firstName;
         @ApiModelProperty(required = true, value = "Last name", example = "User", position = 3)
         private String lastName;
-        @ApiModelProperty(value = "Additional roles", example = "[ROLE_LICENCE_VARY]", dataType = "Set", position = 4)
```
petergphillips: This is never used in the front end so removed

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/maintain/AuthUserService.java
```diff
@@ -64,12 +64,22 @@ public String createUser(final String usernameInput, final String emailInput, fi
         // validate
         validate(username, email, firstName, lastName);
 
+        // get the initial group to assign to - only allowed to be empty if super user
+        final Set<Group> groups;
```
JamieDraperUK: Very subjective but feel like this would look neater if moved to a method.

petergphillips: Agreed, fixed

PR #224, opened 2019-07-15
DT-80 assignable roles for groups

petergphillips:


src/main/java/uk/gov/justice/digital/hmpps/oauth2server/auth/model/GroupAssignableRole.java
```diff
@@ -0,0 +1,37 @@
...
+public class GroupAssignableRole implements Serializable {
+    @Id
+    @ManyToOne
+    @JoinColumn(name = "role_id")
+    private Authority role;
+
+    @Id
+    @ManyToOne
+    @JoinColumn(name = "group_id")
+    private Group group;
```
petergphillips: this is naff but can't work out how to remove this association since it should just from `group` -> `role`

PR #225, opened 2019-07-15
DT-80: Restrict group managers to add / remove only if user / manager…

petergphillips:
… relationship exists


src/test/java/uk/gov/justice/digital/hmpps/oauth2server/maintain/AuthUserRolesServiceTest.java
```diff
@@ -62,9 +63,33 @@ public void addRole_invalidRole() {
                 isInstanceOf(AuthUserRoleException.class).hasMessage("Add role failed for field role with reason: invalid");
     }
 
+    @Test
+    public void addRole_noaccess() {
+        final var user = new UserEmail("user");
+        user.setGroups(Set.of(new Group("group", "desc")));
+        final var role = new Authority("ROLE_LICENCE_VARY", "Role Licence Vary");
```
JamieDraperUK: Feels like this code should either be moved closer to usage or instantiated inline on line 74 to improve legibility. 🤔

PR #232, opened 2019-08-02
DT-127: Modify for new govuk-frontend

petergphillips:


src/main/resources/static/css/app-ie8.css
```diff
@@ -24,7 +24,7 @@
 }
 
 #cancel {
-  background-color: #dee0e2;
+  background-color: #f3f2f1;
```
petergphillips: colours updated according to colours section in https://github.com/alphagov/govuk-frontend/releases/tag/v3.0.0

src/main/resources/templates/changePassword.html
```diff
@@ -26,7 +26,7 @@
   <div
       th:unless="${#arrays.isEmpty(errornew) and #arrays.isEmpty(errorconfirm)}"
       class="govuk-error-summary"
-      aria-labelledby="error-summary-title" role="alert" data-module="error-summary">
+      aria-labelledby="error-summary-title" role="alert" data-module="govuk-error-summary">
```
petergphillips: all `data-module` sections now have to be prefixed by `govuk`

PR #236, opened 2019-08-05
DT-116: Add last logged in and lock accounts if not used for 90 days

petergphillips:


src/main/java/uk/gov/justice/digital/hmpps/oauth2server/timed/DisableInactiveAuthUsers.java
```diff
@@ -0,0 +1,47 @@
...
+            } catch (final Exception e) {
+                // have to catch the exception here otherwise scheduling will stop
+                log.error("Caught exception {} whilst trying to disable users", e.getClass().getSimpleName(), e);
+                errorCount++;
+                telemetryClient.trackEvent("DisableInactiveAuthUsersError", null, null);
+                lastRunFailed = true;
+            }
+
+        } while ((processed > 9 || lastRunFailed) && errorCount < 3);
+
```
timharrison-moj: Perhaps configurable properties for the batch size & error count before quitting the job?

src/main/resources/db/auth/V2_5__drop_authority.sql
```diff
@@ -0,0 +1 @@
+drop table authority;
```
timharrison-moj: I was wondering what this was used for - couldn't see any usage! I will remove from the auth page too.

src/test/java/uk/gov/justice/digital/hmpps/oauth2server/timed/DisableInactiveAuthUsersServiceTest.java
```diff
@@ -0,0 +1,69 @@
...
+    @Before
+    public void setUp() {
+        service = new DisableInactiveAuthUsersService(userEmailRepository, telemetryClient);
+    }
+
+    @Test
+    public void findAndDisableInactiveAuthUsers_Processed() {
+        final var users = List.of(UserEmail.of("user"), UserEmail.of("joe"));
+        when(userEmailRepository.findTop10ByLastLoggedInBeforeAndEnabledIsTrueAndMasterIsTrueOrderByLastLoggedIn(any()))
+                .thenReturn(users);
```
timharrison-moj: Is that repository method name a little long? Meaningful name, but maybe a little on the unwieldy side?

src/main/resources/db/auth/V2_5__drop_authority.sql
petergphillips: :+1: 
The table data was copied over to the new `roles` table in a previous PR, but didn't want to drop the table until the new table had gone live otherwise rolling deployment wouldn't have worked

src/test/java/uk/gov/justice/digital/hmpps/oauth2server/timed/DisableInactiveAuthUsersServiceTest.java
petergphillips: Completely agree, however this is spring boot jpa.  The functionality is determined from the method name (no implementations here), so shortening the method name will remove some of the functionality :grinning:

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/timed/DisableInactiveAuthUsers.java
petergphillips: Batch size is a difficult one, didn't really want to increase the size of the JPA method any more by adding in a parameterised batch size!
Error count could be made configurable, however didn't want to add to properties until we needed them.
If we end up changing either value at all then quite happy to parameterise instead, otherwise it just adds noise...

PR #242, opened 2019-08-14
DT-153: Add in contact us page for each service with an email

petergphillips:


src/main/resources/templates/resetPassword.html
```diff
@@ -1,14 +1,12 @@
...
 <html xmlns:layout="http://www.ultraq.net.nz/thymeleaf/layout" lang="en"
       xmlns:th="http://www.thymeleaf.org"
-      layout:decorate="~{fragments/layout}">
+      layout:decorate="~{fragments/layout}" th:with="showBack=true">
 <head>
   <title>Reset Password</title>
 </head>
 <body>
 <div layout:fragment="content" class="govuk-grid-column-two-thirds">
-  <a th:href="@{/login}" class="govuk-back-link" id="back-link">Back</a>
```
petergphillips: found out that back link needs to be outside main content so skip link works correctly - therefore moved into layout instead

PR #245, opened 2019-08-15
DT-119: Allow group managers to manage email addresses for their users

petergphillips:


src/main/java/uk/gov/justice/digital/hmpps/oauth2server/resource/api/AuthUserController.java
```diff
@@ -223,21 +222,22 @@ private String createInitialPasswordUrl(@ApiIgnore final HttpServletRequest requ
...
+    @PreAuthorize("hasAnyRole('ROLE_MAINTAIN_OAUTH_USERS', 'ROLE_AUTH_GROUP_MANAGER')")
     @ApiOperation(value = "Amend a user.", notes = "Amend a user.", nickname = "amendUser",
             consumes = "application/json", produces = "application/json")
     @ApiResponses(value = {
             @ApiResponse(code = 204, message = "OK"),
             @ApiResponse(code = 400, message = "Bad request e.g. if validation failed or if the amendments are disallowed", response = ErrorDetail.class),
             @ApiResponse(code = 401, message = "Unauthorized.", response = ErrorDetail.class),
-            @ApiResponse(code = 404, message = "User not found.", response = ErrorDetail.class)})
+            @ApiResponse(code = 404, message = "User not found.", response = ErrorDetail.class),
+            @ApiResponse(code = 409, message = "Unable to amend user, the user is not within one of your groups", response = ErrorDetail.class)})
```
JamieDraperUK: Not sure about using HTTP 409 here, due to the broad use cases. This feels more like a security violation. 403 might be a bit more descriptive because the user is authenticated but not authorised to carry out this operation.

petergphillips: Agreed, will change.  In my defence I copied the status from a different endpoint...

PR #251, opened 2019-09-10
DT-187 Added a POM user with document role for dev local

andymarke:


petergphillips:
Already in master so not required

PR #263, opened 2019-10-05
DT-199: Rename user email to user prior to changes

petergphillips:


build.gradle
```diff
@@ -7,7 +7,7 @@ import java.time.format.DateTimeFormatter
 plugins {
     id 'java'
     id 'groovy'
-    id 'org.springframework.boot' version '2.1.8.RELEASE'
+    id 'org.springframework.boot' version '2.1.9.RELEASE'
```
petergphillips: spring boot upgrade :joy:

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/auth/model/User.java
```diff
@@ -70,11 +70,14 @@
     private Set<Authority> authorities = new HashSet<>();
 
     @OneToMany
+    @JoinTable(name = "user_email_groups",
```
petergphillips: had to add explicit name of table here as was wanting to be `user_groups` instead...

PR #265, opened 2019-10-07
DT-199: Revert unnecessary changes and tidy delete user token

petergphillips:


src/main/java/uk/gov/justice/digital/hmpps/oauth2server/auth/repository/UserTokenRepository.java
```diff
@@ -1,15 +1,7 @@
...
 import org.springframework.data.repository.CrudRepository;
-import uk.gov.justice.digital.hmpps.oauth2server.auth.model.User;
 import uk.gov.justice.digital.hmpps.oauth2server.auth.model.UserToken;
-import uk.gov.justice.digital.hmpps.oauth2server.auth.model.UserToken.TokenType;
-
-import java.util.List;
-import java.util.Optional;
 
 public interface UserTokenRepository extends CrudRepository<UserToken, String> {
-    Optional<UserToken> findByTokenTypeAndUser(TokenType tokenType, User user);
```
petergphillips: most access (apart from looking up tokens) should be through `UserRepository` instead

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/auth/model/UserToken.java
```diff
@@ -26,7 +26,7 @@
     @Column(name = "token_expiry", nullable = false)
     private LocalDateTime tokenExpiry;
 
-    @ManyToOne(fetch = FetchType.LAZY)
+    @ManyToOne
```
petergphillips: since going via `UserToken` will always want to get the related user

PR #266, opened 2019-10-07
DT-199: Initial stab at adding user id field

petergphillips:


src/main/resources/db/auth/V3_3__user_token.sql
```diff
@@ -0,0 +1,15 @@
...
+UPDATE user_token ut
+SET user_id = (select user_id FROM users u where u.username = ut.username);
+
+-- change user_id now to be not nullable
+ALTER TABLE user_token
+    MODIFY user_id UNIQUEIDENTIFIER NOT NULL;
+
+-- and username now to be nullable
+ALTER TABLE user_token
+    MODIFY username NULL;
```
petergphillips: `username` field will then be removed in a subsequent release

petergphillips:
I've cunningly renamed the oauth data load to 2.10 so that it is run in before the table changes, hence simulating a db upgrade.  After go live will change data to be 9.x again

src/main/resources/db/auth/V3_4__user_token_data.sql
```diff
@@ -0,0 +1,11 @@
...
+UPDATE user_token
+SET user_id = (select u.user_id FROM users u where u.username = user_token.username);
+
+-- change user_id now to be not nullable
+ALTER TABLE user_token ALTER COLUMN user_id UNIQUEIDENTIFIER NOT NULL;
+
+-- and username now to be nullable
+DROP INDEX user_token_username_idx ON user_token;
+
+ALTER TABLE user_token ALTER COLUMN username varchar(30) NULL;
```
petergphillips: `username` to be removed in future PR

src/main/resources/db/auth/V3_0__users.sql
```diff
@@ -0,0 +1,51 @@
+create table users
+(
+    user_id         UNIQUEIDENTIFIER NOT NULL
+        CONSTRAINT user_id_pk PRIMARY KEY,
+    username        VARCHAR(30)      NOT NULL
+        CONSTRAINT username_uk UNIQUE,
+    password        VARCHAR(100),
+    email           VARCHAR(240),
+    first_name      VARCHAR(50),
```
petergphillips: note that I've moved the `first_name` and `last_name` directly onto the `users` table rather than having a `person` table

PR #267, opened 2019-10-08
DT-269: Add in user_id to JWTs

petergphillips:


src/main/java/uk/gov/justice/digital/hmpps/oauth2server/auth/model/User.java
```diff
@@ -26,6 +27,7 @@
     @GeneratedValue(generator = "UUID")
     @GenericGenerator(name = "UUID", strategy = "org.hibernate.id.UUIDGenerator")
     @Column(name = "user_id", updatable = false, nullable = false)
+    @Type(type = "uuid-char")
```
petergphillips: need to explicity map this otherwise bits are the wrong way round - see https://stackoverflow.com/questions/41651681/different-representation-of-uuid-in-java-hibernate-and-sql-server

PR #274, opened 2019-10-17
DT-292: added telemetryModule to add client_id and user_name as custo…

SimonMitchellMOJ:
…mDimension to app insights


src/test/java/uk/gov/justice/digital/hmpps/oauth2server/config/ClientTrackingTelemetryModuleTest.java
```diff
@@ -0,0 +1,86 @@
...
+
+    @Autowired
+    private ClientTrackingTelemetryModule clientTrackingTelemetryModule;
+
+    @Autowired
+    private JwtAuthHelper jwtAuthHelper;
+
+    @Before
+    public void setup() {
+        ThreadContext.setRequestTelemetryContext(new RequestTelemetryContext(1L));
```
andrewrlee: Is it worth tearing down afterwards?

build.gradle
```diff
@@ -116,6 +116,8 @@ dependencies {
 
   compileOnly 'org.projectlombok:lombok:1.18.8'
 
+  testAnnotationProcessor 'org.projectlombok:lombok:1.18.8'
```
petergphillips: should really match the `implementation` version - i.e. 1.18.10

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/config/ClientTrackingTelemetryModule.java
```diff
@@ -0,0 +1,76 @@
...
+import org.springframework.http.HttpHeaders;
+import org.springframework.security.oauth2.provider.token.store.KeyStoreKeyFactory;
+
+import javax.servlet.ServletRequest;
+import javax.servlet.ServletResponse;
+import javax.servlet.http.HttpServletRequest;
+import java.security.KeyPair;
+
+@Configuration
+public class ClientTrackingTelemetryModule implements WebTelemetryModule, TelemetryModule {
```
petergphillips: think you only need to implement `WebTelemetryModule`?  That's the only one you are implementing

SimonMitchellMOJ: TelemetryModule is required, if its not there the customDimensions are not added to app insights

PR #276, opened 2019-10-18
DT-310: Spring boot 2 upgrade and add owasp

petergphillips:


build.gradle
```diff
@@ -99,10 +100,9 @@ dependencies {
   implementation 'org.springframework.boot:spring-boot-starter-thymeleaf'
   implementation 'org.thymeleaf.extras:thymeleaf-extras-springsecurity5'
   implementation 'nz.net.ultraq.thymeleaf:thymeleaf-layout-dialect:2.4.1'
-  implementation 'org.webjars:bootstrap:4.3.1'
   implementation 'uk.gov.service.notify:notifications-java-client:3.15.0-RELEASE'
 
-  implementation 'org.flywaydb:flyway-core:5.2.4'
+  implementation 'org.flywaydb:flyway-core:6.0.6'
```
petergphillips: can now upgrade flyway :grin:

src/main/resources/application.yml
```diff
@@ -53,7 +53,7 @@ management:
     health:
       cache:
         time-to-live: 2000ms
-      show-details: always
+      show-components: always
```
petergphillips: change to health check output - now `components` instead

PR #281, opened 2019-10-24
DT-140: Add in match subdomains for logout redirect too

petergphillips:


src/test/java/uk/gov/justice/digital/hmpps/oauth2server/resource/RedirectingLogoutSuccessHandlerTest.java
```diff
@@ -31,7 +32,7 @@
 
     @Before
     public void setUp() {
-        redirectingLogoutSuccessHandler = new RedirectingLogoutSuccessHandler(clientDetailsService, "/path");
+        redirectingLogoutSuccessHandler = new RedirectingLogoutSuccessHandler(clientDetailsService, "/path", false);
```
brightonsbox: Would it be possible / valuable to add a test for the matchSubdomains = true case?

petergphillips: good idea, will raise a separate PR for that

PR #288, opened 2019-11-08
DT-355 for testing allow ITAG_USER to migrate keyworkers

Mjwillis:


src/main/resources/db/dev/data/nomis/V9_1__user_data.sql
```diff
@@ -261,6 +261,7 @@ INSERT INTO USER_CASELOAD_ROLES (ROLE_ID, CASELOAD_ID, USERNAME)
 VALUES (-1, 'MDI', 'ITAG_USER'),
   (-1, 'MDI', 'OLD_NOMIS_USER'),
   (1, 'NWEB', 'ITAG_USER'),
+  (4, 'NWEB', 'ITAG_USER'),
```
JamieDraperUK: Are these meant to be unique?

brightonsbox: CONSTRAINT "USER_CASELOAD_ROLES_PK" PRIMARY KEY ("USERNAME", "CASELOAD_ID", "ROLE_ID")

So as long as the combo is unique this is fine

lbennett-moj: the role Id is different

Mjwillis: No, a user can have lots of roles

PR #291, opened 2019-11-19
Sm dt 374 add last logged in to auth user details information

SimonMitchellMOJ:


src/main/java/uk/gov/justice/digital/hmpps/oauth2server/resource/api/AuthUserController.java
```diff
@@ -300,6 +301,9 @@ private String createInitialPasswordUrl(@ApiIgnore final HttpServletRequest requ
         private boolean enabled;
         @ApiModelProperty(required = true, value = "Email address has been verified", example = "false", position = 8)
         private boolean verified;
+        @ApiModelProperty(required = true, value = "Last time user logged in", example = "01/01/2001", position = 9)
```
andrewrlee: is this required? what if the user has never logged in?

PR #293, opened 2019-11-26
DT-367: Tidy user services prior to changes

petergphillips:


src/test/java/uk/gov/justice/digital/hmpps/oauth2server/maintain/AuthUserServiceTest.java
```diff
@@ -315,8 +316,161 @@ public void amendUser_passwordSet() {
                 hasMessageContaining("reason: notinitial");
     }
 
+    @Test
+    public void getAuthUserByUsername() {
```
mikehalmamoj: I think if you do a "git mv" instead of refactoring in Intellij then git doesn't treat this as delete + add so there is less noise in the PR.  (I think it's a bug in Intellij but I'm struggling to recall the details...)

mikehalmamoj: Nope!  Ignore me, totally different thing.

petergphillips: I was about to write a comment until I saw your update :joy:

PR #296, opened 2019-11-28
DT-390 owasp

petergphillips:


build.gradle
```diff
@@ -109,6 +119,7 @@ dependencies {
   implementation 'org.apache.commons:commons-text:1.8'
   implementation 'com.microsoft.sqlserver:mssql-jdbc:7.4.1.jre11'
   implementation 'com.github.timpeeters:spring-boot-graceful-shutdown:2.1.1'
+  implementation 'com.google.guava:guava:28.1-jre'
```
brightonsbox: Did you definitely mean to add these dependencies in now?

petergphillips: well tbh the guava dependency is definitely required, we were picking up v20 without it which would fail the check in the morning :grinning:

brightonsbox: Ah ok, so these were being pulled in transitively before 👍

PR #299, opened 2019-12-02
DT-394 automatic delius email verify

petergphillips:


src/main/java/uk/gov/justice/digital/hmpps/oauth2server/auth/model/User.java
```diff
@@ -93,13 +93,6 @@ public static User of(final String username) {
         return User.builder().username(username).build();
     }
 
-    public static User fromUserPersonDetails(final UserPersonDetails userPersonDetails) {
```
petergphillips: moved so that `UserPersonDetails` converts to a `User` instead

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/security/UserRetriesService.java
```diff
@@ -33,8 +33,14 @@ public void resetRetriesAndRecordLogin(final UserPersonDetails userPersonDetails
 
         // and record last logged in as now too (doing for all users to prevent confusion)
         final var userOptional = userRepository.findByUsername(username);
-        final var user = userOptional.orElseGet(() -> User.fromUserPersonDetails(userPersonDetails));
+        final var user = userOptional.orElseGet(userPersonDetails::toUser);
         user.setLastLoggedIn(LocalDateTime.now());
+
+        // copy across email address on each successful login
+        if (userPersonDetails instanceof DeliusUserPersonDetails) {
```
petergphillips: copy email address over on every successful login

PR #300, opened 2019-12-03
DT-394: Add delius and nomis default roles and fix locked integration

petergphillips:


src/main/java/uk/gov/justice/digital/hmpps/oauth2server/delius/service/DeliusUserService.java
```diff
@@ -91,16 +91,6 @@ public void changePassword(final String username, final String password) {
         restTemplate.postForEntity("/users/{username}/password", requestEntity, Void.class, Map.of("username", username));
     }
 
-    public void lockAccount(final String username) {
-        restTemplate.postForEntity("/users/{username}/lock", null, Void.class, Map.of("username", username));
```
petergphillips: this was actually enabling / disabling user instead

PR #306, opened 2019-12-06
DT-399 map delius roles

petergphillips:


src/main/java/uk/gov/justice/digital/hmpps/oauth2server/Oauth2serverApplication.java
```diff
@@ -2,10 +2,10 @@
 
 import org.springframework.boot.SpringApplication;
 import org.springframework.boot.autoconfigure.SpringBootApplication;
-import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.boot.context.properties.ConfigurationPropertiesScan;
 
 @SpringBootApplication
-@ConfigurationProperties
+@ConfigurationPropertiesScan
```
petergphillips: apparently with spring boot 2.2 need the `Scan` option to then automatically pick up classes with `ConfigurationProperties`

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/auth/model/User.java
```diff
@@ -74,25 +71,26 @@
...
     @OneToMany(mappedBy = "user", cascade = CascadeType.ALL, orphanRemoval = true)
-    @Builder.Default
     @Setter(AccessLevel.NONE)
     private Set<UserToken> tokens = new HashSet<>();
 
     public static User of(final String username) {
         return User.builder().username(username).build();
     }
 
+    public static UserBuilder builder() {
```
petergphillips: delombok required as now called from Kotlin - which is then processed before lombok

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/auth/model/User.java
```diff
@@ -158,4 +156,102 @@ public void removeToken(final UserToken userToken) {
...
+    public static class UserBuilder {
+        private UUID id;
+        private String username;
+        private String password;
+        private String email;
+        private boolean verified;
+        private boolean locked;
+        private boolean enabled;
+        private AuthSource source;
+        private LocalDateTime passwordExpiry = LocalDateTime.now();
```
petergphillips: note that delombok didn't add the defaults to the builder, so had to add manually

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/config/RestTemplateConfiguration.kt
```diff
@@ -40,7 +41,11 @@ open class RestTemplateConfiguration(private val apiDetails: ClientCredentialsRe
 
 }
 
-@Suppress("DEPRECATION")
-@Configuration
```
petergphillips: don't need the `Configuration` annotation under spring 2.2.1 with `Scan`

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/delius/service/DeliusUserService.kt
```diff
@@ -0,0 +1,98 @@
...
+      DeliusUserPersonDetails(
+          firstName = userDetails.firstName,
+          surname = userDetails.surname,
+          roles = mapUserRolesToAuthorities(userDetails.roles),
+          username = username,
+          email = userDetails.email,
+          enabled = userDetails.enabled)
+
+  private fun mapUserRolesToAuthorities(userRoles: List<UserRole>): Collection<GrantedAuthority> =
+      userRoles.map { (name) -> mappings.mappings[name] }
```
petergphillips: notice the `[]` syntax for retrieving an element from a list

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/delius/service/DeliusUserService.kt
```diff
@@ -0,0 +1,98 @@
...
+          .filterNotNull()
+          .flatMap { r -> r.map(::SimpleGrantedAuthority) }
```
petergphillips: the `::` is the same as in java - calls the constructor with the single argument

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/delius/model/DeliusUserPersonDetails.kt
```diff
@@ -0,0 +1,45 @@
...
+
+  override fun getAuthSource(): String = "delius"
+
+  override fun toUser(): User =
+      User.builder().username(username).source(AuthSource.delius).email(email).verified(true).build()
+
+  override fun eraseCredentials() {}
+
+  // add in ROLE_PROBATION to standard roles
+  override fun getAuthorities(): Collection<GrantedAuthority?> = roles.plus(SimpleGrantedAuthority("ROLE_PROBATION"))
```
petergphillips: note that in kotlin land don't need to play around with creating stream of two sets, can just call `plus` on an immutable collection to create a new list with that element added

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/delius/service/DeliusUserService.kt
andrewrlee: was a bit confused but its a map not a list right?

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/auth/model/User.java
Mjwillis: It a shame we need to do this

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/delius/service/DeliusUserService.kt
andrewrlee: Each of these map, filter, flatmaps iterates the collection again doesn't it? unlike java streams which would do  the iteration once.
These collections being tiny, makes it a moot point  - but just for my understanding?

src/main/resources/application.yml
```diff
@@ -116,6 +116,11 @@ delius:
     access-token-uri: ${auth.token.url}
     authentication-scheme: header
     client-authentication-scheme: header
+  roles:
+    mappings:
+      # couldn't work out how to provide empty map here, so dummy values here
+      ROLE_DUMMY:
+        - 'ROLE_DUMMY'
```
andrewrlee: meh.. `ROLE_DUMMY: {}` would work wouldn't it? 

yaml being a superset of JS

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/delius/model/DeliusUserPersonDetails.kt
```diff
@@ -0,0 +1,45 @@
...
+import org.springframework.security.core.authority.SimpleGrantedAuthority
+import uk.gov.justice.digital.hmpps.oauth2server.auth.model.User
+import uk.gov.justice.digital.hmpps.oauth2server.security.AuthSource
+import uk.gov.justice.digital.hmpps.oauth2server.security.UserPersonDetails
+
+data class DeliusUserPersonDetails(private val surname: String,
+                                   private val firstName: String,
+                                   private val username: String,
+                                   private val enabled: Boolean = false,
+                                   val email: String,
```
mikehalmamoj: Should this come before the optional parameters?  Or does it not matter because you use named parameters?

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/delius/service/DeliusUserService.kt
petergphillips: oops you're right :man_facepalming: I meant map

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/delius/model/DeliusUserPersonDetails.kt
petergphillips: good spot, you're right.

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/delius/service/DeliusUserService.kt
petergphillips: agreed, meant to use `Sequences` if we want the equivalent lazy behaviour

petergphillips: however just realised that can use `mapNotNull` which reduces it down a bit more

src/main/resources/application.yml
petergphillips: removed, will add default to deploy script instead

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/auth/model/User.java
petergphillips: agreed, but it's only temporary until we migrate the class to a kotlin `data class`

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/delius/model/DeliusUserPersonDetails.kt
petergphillips: now fixed

.circleci/config.yml
```diff
@@ -1,7 +1,7 @@
 defaults: &defaults
   working_directory: ~/app
   environment:
-    JAVA_TOOL_OPTIONS: -Xmx1024m -XX:ParallelGCThreads=2 -XX:ConcGCThreads=2 -XX:ParallelGCThreads=2 -Djava.util.concurrent.ForkJoinPool.common.parallelism=2
+    JAVA_TOOL_OPTIONS: -Xmx512m -XX:ParallelGCThreads=2 -XX:ConcGCThreads=2 -XX:ParallelGCThreads=2 -Djava.util.concurrent.ForkJoinPool.common.parallelism=2
```
petergphillips: decrease memory usage in hope that OOM killer doesn't kill our java process :crossed_fingers:

PR #307, opened 2019-12-06
DT-407 fix delius reset password

petergphillips:


src/main/java/uk/gov/justice/digital/hmpps/oauth2server/delius/service/DeliusUserService.kt
```diff
@@ -73,10 +73,11 @@ open class DeliusUserService(@Qualifier("deliusApiRestTemplate") private val res
 
   private fun mapUserDetailsToDeliusUser(userDetails: UserDetails, username: String): DeliusUserPersonDetails =
       DeliusUserPersonDetails(
-          username = username,
+          username = username.toUpperCase(),
```
petergphillips: usernames always uppercase in auth

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/delius/service/DeliusUserService.kt
```diff
@@ -73,10 +73,11 @@ open class DeliusUserService(@Qualifier("deliusApiRestTemplate") private val res
...
+          userId = userDetails.userId,
           firstName = userDetails.firstName,
           surname = userDetails.surname,
-          email = userDetails.email,
+          email = userDetails.email.toLowerCase(),
```
petergphillips: and emails always lowercase

PR #308, opened 2019-12-09
DT-408: Lock delius users if password incorrect

petergphillips:


src/main/java/uk/gov/justice/digital/hmpps/oauth2server/security/DeliusUserDetailsService.kt
```diff
@@ -15,11 +15,18 @@ import uk.gov.justice.digital.hmpps.oauth2server.delius.service.DeliusUserServic
...
-open class DeliusUserDetailsService(private val deliusUserService: DeliusUserService) :
+open class DeliusUserDetailsService(private val deliusUserService: DeliusUserService,
+                                    private val userService: UserService) :
     UserDetailsService, AuthenticationUserDetailsService<PreAuthenticatedAuthenticationToken> {
 
-  override fun loadUserByUsername(username: String): UserDetails =
-      deliusUserService.getDeliusUserByUsername(username).orElseThrow { UsernameNotFoundException(username) }
+  override fun loadUserByUsername(username: String): UserDetails {
+    // need to check first that the user hasn't been locked in auth - as we are handling locking on behalf of delius
+    val locked = userService.findUser(username).map { !it.isAccountNonLocked }.orElse(false)
```
petergphillips: have to look up user in both auth and delius

src/test/resources/mappings/authenticate_failure.json
```diff
@@ -5,9 +5,9 @@
     "bodyPatterns": [
       {
         "equalToJson": {
-          "username": "delfail",
-          "ignoreExtraElements": true
```
petergphillips: this was wrong and never matched

PR #310, opened 2019-12-10
DT-388 support link

mikehalmamoj:
Get the support link for the initial email template from the group of the user being created.

Currently the only requirement is for PECS users to use the BOOK_NOW service  so we have hardcoded for any group starting with "PECS" to grab the support link from the BOOK_NOW service.  If this requirement is extended to other users we probably need a new table to help make this decision.


src/test/java/uk/gov/justice/digital/hmpps/oauth2server/maintain/AuthUserServiceTest.java
```diff
@@ -122,19 +126,23 @@ public void createUser_emailValidation() throws VerifyEmailException {
 
     @Test
     public void createUser_successLinkReturned() throws VerifyEmailException, CreateUserException, NotificationClientException {
+        mockServiceOfNameWithSupportLink("NOMIS", "nomis_support_link");
```
mikehalmamoj: I'm not overjoyed about the additional mock required for all existing createUser/amendUser calls...

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/maintain/AuthUserService.java
```diff
@@ -111,7 +115,16 @@ public String createUser(final String usernameInput, final String emailInput, fi
...
                 .authorities(roles)
                 .groups(groups).build();
-        return saveAndSendInitialEmail(url, user, creator, "AuthUserCreate");
+
+        return saveAndSendInitialEmail(url, user, creator, "AuthUserCreate", groups);
+    }
+
+    private String getInitialEmailSupportLink(Collection<Group> groups) {
+        String serviceCode = "NOMIS";
+        if (groups.stream().anyMatch(group -> group.getGroupCode().startsWith("PECS"))) {
```
mikehalmamoj: If a group code starts with PECS it is considered a PECS user.  I wonder if this should be contains rather than startsWith?

Mjwillis: Starts with is fine for now

src/test/groovy/uk/gov/justice/digital/hmpps/oauth2server/api/specs/AuthUserSpecification.groovy
```diff
@@ -20,7 +20,7 @@ class AuthUserSpecification extends TestSpecification {
 
     def 'Create User endpoint succeeds to create user data'() {
         def username = RandomStringUtils.randomAlphanumeric(10)
-        def user = [email: 'bob@bobdigital.justice.gov.uk', firstName: 'Bob', lastName: 'Smith'] as NewUser
+        def user = [email: 'bob@bobdigital.justice.gov.uk', firstName: 'Bob', lastName: 'Smith', groupCode: 'SITE_1_GROUP_1'] as NewUser
```
petergphillips: shouldn't need this - test should work without it

src/test/java/uk/gov/justice/digital/hmpps/oauth2server/maintain/AuthUserServiceTest.java
petergphillips: agreed, sort of pollutes rest of test.  Default should just be set in the `setUp` method and then can be explicitly overridden

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/maintain/AuthUserService.java
```diff
@@ -111,7 +115,16 @@ public String createUser(final String usernameInput, final String emailInput, fi
                 .person(person)
                 .authorities(roles)
                 .groups(groups).build();
-        return saveAndSendInitialEmail(url, user, creator, "AuthUserCreate");
+
+        return saveAndSendInitialEmail(url, user, creator, "AuthUserCreate", groups);
+    }
+
+    private String getInitialEmailSupportLink(Collection<Group> groups) {
+        String serviceCode = "NOMIS";
```
petergphillips: really don't like unnecessary mutable variable here

petergphillips: since you're streaming anyway can be replaced with something like
```        
final var serviceCode = groups.stream().map(Group::getGroupCode).filter(g -> g.startsWith("PECS")).map(g -> "BOOK_NOW").findFirst().orElse("NOMIS");
```

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/maintain/AuthUserService.java
```diff
@@ -111,7 +115,16 @@ public String createUser(final String usernameInput, final String emailInput, fi
                 .person(person)
                 .authorities(roles)
                 .groups(groups).build();
-        return saveAndSendInitialEmail(url, user, creator, "AuthUserCreate");
+
+        return saveAndSendInitialEmail(url, user, creator, "AuthUserCreate", groups);
+    }
+
+    private String getInitialEmailSupportLink(Collection<Group> groups) {
```
petergphillips: prefer all variables and parameters to be `final`

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/maintain/AuthUserService.java
```diff
@@ -111,7 +115,16 @@ public String createUser(final String usernameInput, final String emailInput, fi
...
+        String serviceCode = "NOMIS";
+        if (groups.stream().anyMatch(group -> group.getGroupCode().startsWith("PECS"))) {
+            serviceCode = "BOOK_NOW";
+        }
+        return oauthServiceRepository.findById(serviceCode).map(service -> service.getEmail()).orElseThrow();
```
petergphillips: can be replaced with lambda `Service::getEmail`

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/maintain/AuthUserService.java
```diff
@@ -129,17 +142,19 @@ public String createUser(final String usernameInput, final String emailInput, fi
         return Optional.of(authUserGroups.stream().filter(g -> g.getGroupCode().equals(groupCode)).findFirst().orElseThrow(() -> new CreateUserException("groupCode", "notfound")));
     }
 
-    private String saveAndSendInitialEmail(final String url, final User user, final String creator, final String eventPrefix) throws NotificationClientException {
+    private String saveAndSendInitialEmail(final String url, final User user, final String creator, final String eventPrefix, Collection<Group> groups) throws NotificationClientException {
```
petergphillips: again param can be `final`

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/maintain/AuthUserService.java
mikehalmamoj: It could, but as it is the second imported class called Service it looks like this:
`        return oauthServiceRepository.findById(serviceCode).map(uk.gov.justice.digital.hmpps.oauth2server.auth.model.Service::getEmail).orElseThrow();
`
Which I find less readable.

petergphillips: completely agree, I would normally have therefore imported this one and made the other one fully qualified instead.  Obviously the better solution would be to convert to kotlin and use `import as` instead...

mikehalmamoj: I read that import as has been a feature request in Java since 1998... come on Java.

PR #313, opened 2019-12-13
DT-417: Reduce delius timeout to 5 seconds

petergphillips:


src/main/java/uk/gov/justice/digital/hmpps/oauth2server/config/RestTemplateConfiguration.kt
```diff
@@ -16,17 +15,17 @@ import java.time.Duration
...
-    RootUriTemplateHandler.addTo(deliusApiRestTemplate, "${deliusEndpointUrl}/secure")
-
-    return deliusApiRestTemplate
-  }
+  open fun deliusRestTemplate(restTemplateBuilder: RestTemplateBuilder): OAuth2RestTemplate =
+      restTemplateBuilder
+          .rootUri("${deliusEndpointUrl}/secure")
+          .setConnectTimeout(apiTimeout)
+          .setReadTimeout(apiTimeout)
+          .configure(OAuth2RestTemplate(apiDetails))
```
petergphillips: turns out rest template builder can `configure` templates as well as `build`ing them :grinning:

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/config/RestTemplateConfiguration.kt
```diff
@@ -16,17 +15,17 @@ import java.time.Duration
 @Suppress("DEPRECATION", "SpringJavaInjectionPointsAutowiringInspection")
 @Configuration
 open class RestTemplateConfiguration(private val apiDetails: DeliusClientCredentials,
-                                     @param:Value("\${delius.endpoint.url}") private val deliusEndpointUrl: @URL String,
-                                     @param:Value("\${delius.health.timeout:1s}") private val healthTimeout: Duration) {
+                                     @Value("\${delius.endpoint.url}") private val deliusEndpointUrl: @URL String,
+                                     @Value("\${delius.health.timeout:1s}") private val healthTimeout: Duration,
+                                     @Value("\${delius.endpoint.timeout:5s}") private val apiTimeout: Duration) {
```
petergphillips: @andymarke default timeout set here of 5 seconds, but can be overridden by setting environment variable if we so choose

PR #319, opened 2020-01-06
DT-423: Move create token into token service

petergphillips:


src/main/java/uk/gov/justice/digital/hmpps/oauth2server/auth/model/UserToken.java
```diff
@@ -43,13 +43,32 @@ public boolean hasTokenExpired() {
         return tokenExpiry.isBefore(LocalDateTime.now());
     }
 
-    @Getter
+    public String getToken() {
```
petergphillips: needed for kotlin interoperability

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/auth/model/UserToken.java
```diff
@@ -43,13 +43,32 @@ public boolean hasTokenExpired() {
...
+    public User getUser() {
+        return this.user;
+    }
+
     @AllArgsConstructor
     public enum TokenType {
-        RESET("Reset"),
-        VERIFIED("Verified"),
-        CHANGE("Change");
+        RESET("ResetPassword"),
```
petergphillips: MFA won't be a password token type, so move `password` here instead for other types

PR #321, opened 2020-01-08
DT-423: Add in mfa page

petergphillips:


src/main/java/uk/gov/justice/digital/hmpps/oauth2server/service/MfaService.kt
```diff
@@ -38,17 +36,19 @@ open class MfaService(@Value("\${application.authentication.mfa.whitelist}") whi
...
 
     // 3. Create email with both login and mfa token
 
     // 4. Send email
     return token
   }
 
-  open fun validateToken(code: String): Optional<String> {
+  open fun validateMfaCode(code: String?): Optional<String> {
+    if (code.isNullOrBlank()) return Optional.of("missingcode");
```
petergphillips: nice bit of kotlin here - even though `code` is `null` we can still call `isNullOrBlank` on it :astonished:

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/auth/model/UserToken.java
```diff
@@ -31,14 +33,24 @@
     private User user;
 
     UserToken(final TokenType tokenType, final User user) {
-        this.token = UUID.randomUUID().toString();
+        this.token = tokenType == TokenType.MFA_CODE ? generateIntAsString() : UUID.randomUUID().toString();
```
petergphillips: for mfa codes generate a 6 digit number instead

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/auth/model/UserToken.java
```diff
@@ -31,14 +33,24 @@
...
         this.tokenType = tokenType;
         this.user = user;
 
         final var now = LocalDateTime.now();
         this.tokenExpiry = tokenType == TokenType.CHANGE || tokenType == TokenType.MFA ? now.plusMinutes(20) : now.plusDays(1);
     }
 
+    private String generateIntAsString() {
+        try {
+            final var random = SecureRandom.getInstance("DRBG").nextInt(1000000);
```
petergphillips: Apparently there is `DRBG` since java 9 which is the most secure

src/test/groovy/uk/gov/justice/digital/hmpps/oauth2server/integration/specs/MfaSpecification.groovy
```diff
@@ -41,11 +39,18 @@ class MfaSpecification extends GebReportingSpec {
...
-//        enterCode '123456'
+    then: 'I am redirected to the mfa page'
+    at MfaPage
+
+    when: "I don't enter a code"
+    submitCode " "
+
+    then: "I am shown an error message"
+    at MfaErrorPage
+    errorText == 'Enter the code received in the email'
```
mikehalmamoj: This feels like it should be a separate test?

petergphillips: Yes, probably.  If it was a unit test then I would have separated it out from the start, however the browser tests are a lot slower to setup - in this case it relies on successful login previously etc.  I'll address in subsequent PR.

PR #323, opened 2020-01-09
DT-423 mfa send email

petergphillips:


src/main/java/uk/gov/justice/digital/hmpps/oauth2server/auth/model/User.java
```diff
@@ -176,6 +175,62 @@ public String toString() {
                 '}';
     }
 
+    public UUID getId() {
```
petergphillips: needed for kotlin interoperability

PR #325, opened 2020-01-10
DT-423: Add in verification of mfa code

petergphillips:


src/main/java/uk/gov/justice/digital/hmpps/oauth2server/resource/MfaController.kt
```diff
@@ -46,24 +49,20 @@ open class MfaController(private val jwtAuthenticationSuccessHandler: JwtAuthent
...
-    }
-
     // can just grab token here as validated above
     val username = tokenService.getToken(TokenType.MFA, token).map { it.user.username }.orElseThrow()
 
     // now load the user
     val userPersonDetails = userService.findMasterUserPersonDetails(username).orElseThrow()
-    val successToken = UsernamePasswordAuthenticationToken(userPersonDetails, "code")
 
-    // now remove tokens to ensure it can't be used again
```
petergphillips: moved to the service

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/security/UserStateAuthenticationFailureHandler.kt
```diff
@@ -47,10 +47,10 @@ class UserStateAuthenticationFailureHandler(private val tokenService: TokenServi
       is MfaRequiredException -> {
         // need to break out to perform mfa for the user
         val username = StringUtils.trim(request.getParameter("username").toUpperCase())
-        val mfaPair = mfaService.createTokenAndSendEmail(username)
+        val (token, code) = mfaService.createTokenAndSendEmail(username)
```
petergphillips: destructuring goodness

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/service/MfaService.kt
```diff
@@ -49,24 +51,37 @@ open class MfaService(@Value("\${application.authentication.mfa.whitelist}") whi
...
   }
 
-  open fun validateMfaCode(code: String?): Optional<String> {
-    if (code.isNullOrBlank()) return Optional.of("missingcode");
-    // 1. look up mfa code
-    // 2. fail if expired or invalid - incorrect password attempts?
-    // 3. return empty if okay
-    return Optional.empty()
+  @Transactional(transactionManager = "authTransactionManager")
+  open fun validateAndRemoveMfaCode(token: String, code: String?): Optional<String> {
```
mikehalmamoj: Any good reason not make the return String nullable instead of Optional?

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/service/MfaService.kt
```diff
@@ -49,24 +51,37 @@ open class MfaService(@Value("\${application.authentication.mfa.whitelist}") whi
...
-  open fun resendMfaCode(token: String) {
-    // 1. look up mfa token and find user
-    // 2. Find mfa code for user
-    // 3. send email to user again with token and code
+  open fun resendMfaCode(token: String): String? {
+    val userToken = tokenService.getToken(TokenType.MFA, token).orElseThrow()
+
+    val code = userToken.user.tokens.filter { it.tokenType == TokenType.MFA_CODE }.map { it.token }.firstOrNull()
+
+    code?.run { emailCode(userToken.user, code) }
```
mikehalmamoj: Is this automatically converted to an Optional?  How does that work?

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/service/MfaService.kt
petergphillips: I was chatting to Jamie about this yesterday.  Not really happy with Kotlin having to return`null` instead of `Optional`, since we use `Optional` instead of `null` in java...

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/service/MfaService.kt
petergphillips: no, that is a kotlin `String?` type that you can therefore call things like `isNullOrEmpty` on.  No automatic translation.

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/service/MfaService.kt
petergphillips: Can have another look next week!

petergphillips: I'm not that keen about the `Optional` stuff in the first place (even though I added it all).  With hindsight would be better to just throw an exception and then catch it, would be clearer what is going on as in some cases the optional empty is success, whereas in other cases it is failure...

PR #327, opened 2020-01-14
DT-477 mfa increment retries

petergphillips:


src/main/java/uk/gov/justice/digital/hmpps/oauth2server/auth/model/User.java
```diff
@@ -18,7 +21,6 @@
 
 @Entity
 @Table(name = "USERS")
-@Setter
```
petergphillips: needed for kotlin interoperability

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/auth/model/UserRetries.kt
```diff
@@ -0,0 +1,25 @@
...
+
+@Entity
+@Table(name = "USER_RETRIES")
+data class UserRetries(
+    @Id
+    @Column(name = "username", nullable = false)
+    val username: String? = null) {
+
+  @Column(name = "retry_count")
+  var retryCount: Int = 0
```
petergphillips: by specifying `retryCount` outside the primary constructor kotlin won't include it in the auto generated hashcode and equals (same as the java version)

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/auth/model/UserRetries.kt
```diff
@@ -0,0 +1,25 @@
...
+import javax.persistence.Entity
+import javax.persistence.Id
+import javax.persistence.Table
+
+@Entity
+@Table(name = "USER_RETRIES")
+data class UserRetries(
+    @Id
+    @Column(name = "username", nullable = false)
+    val username: String? = null) {
```
petergphillips: we have to make `username` nullable as `@Entity` requires a no args constructor. Kotlin will generate no args and single args constructor for the field.

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/security/LockingAuthenticationProvider.java
```diff
@@ -91,15 +88,11 @@ private void checkPasswordWithAccountLock(final UserPersonDetails userDetails, f
                 throw new MfaUnavailableException("MFA required, but no email set");
             }
         } else {
-            final var newRetryCount = userRetriesService.incrementRetries(username);
+            final var locked = userRetriesService.incrementRetriesAndLockAccountIfNecessary(userDetails);
```
petergphillips: moved to single method for simplicity as both always needed to be called

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/service/MfaService.kt
```diff
@@ -56,17 +57,26 @@ open class MfaService(@Value("\${application.authentication.mfa.whitelist}") whi
     return Pair(token, code)
   }
 
-  @Transactional(transactionManager = "authTransactionManager")
-  open fun validateAndRemoveMfaCode(token: String, code: String?): Optional<String> {
-    if (code.isNullOrBlank()) return Optional.of("missingcode")
+  @Transactional(transactionManager = "authTransactionManager", noRollbackFor = [LoginFlowException::class, MfaFlowException::class])
```
petergphillips: @mikehalmamoj I've switched from `Optional` to exception processing, see what you think.
Had to therefore exclude the exceptions from the rollback here makes it a bit messier

src/test/java/uk/gov/justice/digital/hmpps/oauth2server/security/LockingAuthenticationProviderTest.kt
```diff
@@ -7,15 +7,12 @@ import com.nhaarman.mockito_kotlin.whenever
...
 import org.junit.Test
-import org.junit.runner.RunWith
-import org.mockito.junit.MockitoJUnitRunner
 import org.springframework.security.authentication.UsernamePasswordAuthenticationToken
 import org.springframework.security.crypto.bcrypt.BCryptPasswordEncoder
 import uk.gov.justice.digital.hmpps.oauth2server.security.LockingAuthenticationProvider.MfaRequiredException
 import uk.gov.justice.digital.hmpps.oauth2server.security.LockingAuthenticationProvider.MfaUnavailableException
 import uk.gov.justice.digital.hmpps.oauth2server.service.MfaService
 
-@RunWith(MockitoJUnitRunner::class)
```
petergphillips: unnecessary in kotlin tests with `mock()`

src/test/groovy/uk/gov/justice/digital/hmpps/oauth2server/integration/specs/MfaSpecification.groovy
```diff
@@ -97,6 +96,45 @@ class MfaSpecification extends GebReportingSpec {
     errorText == 'Email code is incorrect. Please check your email and try again. You will be locked out if you enter the wrong code 3 times.'
   }
 
+  def "MFA user gets locked after 3 invalid attempts"() {
```
andymarke: Is it worth having a test where we get the TFA code wrong, then password wrong then TFA code wrong again just to show there is a global lock count?

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/auth/model/UserRetries.kt
mikehalmamoj: The above has a 4 char indent.

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/service/MfaService.kt
```diff
@@ -56,17 +57,26 @@ open class MfaService(@Value("\${application.authentication.mfa.whitelist}") whi
...
+        if (locked) throw LoginFlowException("locked")
+      }
+      throw it
     }
-    return errors
+
+    tokenService.removeToken(TokenType.MFA, token)
+    tokenService.removeToken(TokenType.MFA_CODE, code)
+
+    if (!userPersonDetails.isAccountNonLocked) throw LoginFlowException("locked")
```
mikehalmamoj: Maybe avoid the double negative?

petergphillips: I know it is a bit naff, but that method comes from `org.springframework.security.core.userdetails.UserDetails`.  Going down the negated version route would mean I would have to create a default method for it in `uk.gov.justice.digital.hmpps.oauth2server.security.UserPersonDetails`, which seems a bit overkill and would expose a method that we only need for the single use case.
The other alternative is to do
```
    if (userPersonDetails.isAccountNonLocked) return

    throw LoginFlowException("locked")
```
which in my mind is harder to read as the success flow then terminates early (unlike the rest of the method where the error flows cause the exception to be thrown.

src/test/java/uk/gov/justice/digital/hmpps/oauth2server/service/MfaServiceTest.kt
```diff
@@ -55,29 +58,50 @@ class MfaServiceTest {
...
-    assertThat(service.validateAndRemoveMfaCode("", "somecode")).get().isEqualTo("someproblem")
+    assertThatThrownBy { service.validateAndRemoveMfaCode("", "somecode") }.isInstanceOf(MfaFlowException::class.java).withFailMessage("someproblem")
   }
 
   @Test
-  fun `validateMfaCode success`() {
-    assertThat(service.validateAndRemoveMfaCode("sometoken", "somecode")).isEmpty
+  fun `validateAndRemoveMfaCode success`() {
+    val userToken = User.of("user").createToken(TokenType.MFA)
+    whenever(tokenService.getToken(any(), anyString())).thenReturn(Optional.of(userToken))
```
mikehalmamoj: Seeing as we are passing 2 strings into the method, should we test that this value is "sometoken" rather than "anystring()"?

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/auth/model/UserRetries.kt
petergphillips: indeed.  It is part of the constructor rather than the method body :grinning:

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/service/MfaService.kt
mikehalmamoj: Why is the rollback cancelled for the MfaFlowException?  Looks like it's just the incrementRetriesAndLock... that needs to happen?  Anyway, I think runtime exceptions are better where "fail fast" makes sense and you can use them with a ControllerAdvice, but in this situation failure requires a side effect (locking the account) so it's going to be messy either way.

src/test/java/uk/gov/justice/digital/hmpps/oauth2server/service/MfaServiceTest.kt
petergphillips: I'll add an extra test or two to `verify` that we're calling the correct methods in next PR

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/auth/model/UserRetries.kt
mikehalmamoj: Ha!

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/service/MfaService.kt
petergphillips: When the user gets the mfa code incorrect we don't need to lock the account, but have incremented the retries i.e. we drop down to the `throw it` in
```java
      if (it.error.equals("invalid")) {
        val locked = userRetriesService.incrementRetriesAndLockAccountIfNecessary(userPersonDetails)
        if (locked) throw LoginFlowException("locked")
      }
      throw it
```

PR #328, opened 2020-01-14
DT-477: Bump libraries to latest

petergphillips:


src/main/resources/application.yml
```diff
@@ -1,7 +1,7 @@
 server:
   servlet:
     context-path: /auth
-  use-forward-headers: true
+  forward-headers-strategy: native
```
petergphillips: old property deprecated, intellij suggestion to use new property instead

src/main/resources/application.yml
```diff
@@ -124,5 +124,3 @@ delius:
 logging.uris.exclude.regex: /auth/health|/auth/info|/auth/ping|/auth/swagger.*|/auth/webjars/.*|/auth/v2/api-docs
 
 graceful.shutdown.enabled: true
-
-azure.application-insights.web.enable-W3C: true
```
petergphillips: not needed anymore since we are using the `AI-Agent.xml` that explicitly defines property instead

PR #329, opened 2020-01-14
DT-477 mfa reset retries

petergphillips:


src/main/java/uk/gov/justice/digital/hmpps/oauth2server/service/MfaService.kt
```diff
@@ -64,19 +64,21 @@ open class MfaService(@Value("\${application.authentication.mfa.whitelist}") whi
     val userToken = tokenService.getToken(TokenType.MFA, token).orElseThrow()
     val userPersonDetails = userService.findMasterUserPersonDetails(userToken.user.username).orElseThrow()
 
+    if (!userPersonDetails.isAccountNonLocked) throw LoginFlowException("locked")
+
     val errors = tokenService.checkToken(TokenType.MFA_CODE, code)
-    errors.map { MfaFlowException(it) }.ifPresent {
-      if (it.error.equals("invalid")) {
+    errors.ifPresent {
+      if (it == "invalid") {
```
petergphillips: kotlin is more like javascript for `==` in that `===` does referential equality and `==` does data equality.  `.equals` isn't therefore needed (apart from bizarre float and double cases...)

PR #334, opened 2020-01-17
DT-479: Track failure to login later in process to prevent duplicates

petergphillips:


src/main/java/uk/gov/justice/digital/hmpps/oauth2server/security/MissingCredentialsException.java
```diff
@@ -1,8 +1,8 @@
 package uk.gov.justice.digital.hmpps.oauth2server.security;
 
-import org.springframework.security.core.AuthenticationException;
+import org.springframework.security.authentication.AccountStatusException;
 
-class MissingCredentialsException extends AuthenticationException {
+class MissingCredentialsException extends AccountStatusException {
```
petergphillips: account status exception will prevent us retrying all authentication providers.  In the missing credentials case there is absolutely no point trying all three.

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/security/UserStateAuthenticationFailureHandler.kt
```diff
@@ -57,21 +61,36 @@ class UserStateAuthenticationFailureHandler(private val tokenService: TokenServi
...
+            Pair("missinguser", null)
+          }
+        } else {
+          Pair("missingpass", null)
         }
       }
-      is DeliusAuthenticationServiceException -> builder.add("invalid").add("deliusdown")
-      else -> builder.add("invalid")
+      is DeliusAuthenticationServiceException -> Pair("invalid", "deliusdown")
+      else -> Pair("invalid", null)
```
mikehalmamoj: That's a few nulls... better to do listOf?

src/test/java/uk/gov/justice/digital/hmpps/oauth2server/resource/MfaControllerTest.kt
```diff
@@ -109,6 +109,15 @@ class MfaControllerTest {
     assertThat(modelAndView).isNull()
   }
 
+  @Test
+  fun `mfaChallenge success telemetry`() {
+    val user = User.of("someuser")
+    whenever(tokenService.getToken(any(), anyString())).thenReturn(Optional.of(UserToken("otken", TokenType.MFA, null, user)))
```
mikehalmamoj: typo

petergphillips: deliberate - I try not to create inputs with the same name as the parameter :grin:

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/security/UserStateAuthenticationFailureHandler.kt
petergphillips: Not sure.  I initially had it as a `listOf` but had a problem in that it should really be two parameters - one required and the other optional.  I therefore had to do a `first` with a throw and looked a bit worse - I'll raise another PR to explain better

src/test/java/uk/gov/justice/digital/hmpps/oauth2server/resource/MfaControllerTest.kt
mikehalmamoj: Ah, interesting.  On my last project we got quite anal about the intent of test data - "some token" would indicate a value to be tested and "any token" would indicate that we don't care about the value in this test.  Which accidentally makes them a different name to the parameter.

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/security/UserStateAuthenticationFailureHandler.kt
mikehalmamoj: I don't really have a strong opinion either way, just wondered about your rationale.

PR #341, opened 2020-01-22
DT-383 auth user name changes 2

petergphillips:


src/test/groovy/uk/gov/justice/digital/hmpps/oauth2server/integration/specs/UserDetailsSpecification.groovy
```diff
@@ -0,0 +1,87 @@
...
+    loginAs AUTH_ADM, 'password123456'
+
+    when: 'I navigate to the user details page'
+    at HomePage
+    def currentName = principalName
+    to UserDetailsPage
+
+    then: 'The user details page is displayed with existing name'
+    at UserDetailsPage
+    "${firstNameInput} ${lastNameInput}" == currentName
```
andymarke: Very minor, but personally I find hardcoding the names make the test read better, (same for randomLastName below).  Feel free to ignore this though. Both methods 100% valid

petergphillips: Good plan.  I'll remove the random string utils and instead hardcode.  Can't remove the `currentName` stuff though as it depends what value is currently set for the user on the home page - since there are 3 tests all manipulating the same user would be better to not hardcode and make brittle

PR #343, opened 2020-01-22
NN-2392 add new role and assign it to court user

JamieDraperUK:


src/main/resources/db/auth/V3_14__external_appointment_user_group.sql
```diff
@@ -0,0 +1,2 @@
+INSERT INTO groups (group_id, group_code, group_name)
+VALUES (newid(), 'EXTERNAL_APPOINTMENT', 'External appointment');
```
petergphillips: should be `VIDEO_LINK_COURT_USER`

src/main/resources/db/auth/V3_13__add_global_appointment_user_role_and_assign_to_court_user.sql
```diff
@@ -0,0 +1,8 @@
+INSERT INTO roles (role_id, role_code, role_name)
+VALUES (newid(), 'GLOBAL_APPOINTMENT', 'global appointment');
```
petergphillips: Please could you capitalise both _Global_ and _Appointment_ to keep consistent - e.g. see V2_2__user_roles.sql

PR #345, opened 2020-01-23
DT-383: Fixes for issues discovered during testing

petergphillips:


src/main/java/uk/gov/justice/digital/hmpps/oauth2server/maintain/AuthUserService.kt
```diff
@@ -180,32 +180,28 @@ open class AuthUserService(private val userRepository: UserRepository,
...
     verifyEmailService.validateEmailAddress(email)
   }
 
   @Throws(CreateUserException::class)
   private fun validate(firstName: String?, lastName: String?) {
-    if (StringUtils.length(firstName) < MIN_LENGTH_FIRST_NAME) {
-      throw CreateUserException("firstName", "length")
+    if (firstName.isNullOrBlank()) throw CreateUserException("firstName", "required")
+    else {
+      if (firstName.contains('<') || firstName.contains('>')) throw CreateUserException("firstName", "invalid")
```
petergphillips: note that even though `firstName` is nullable, we can safely use `.` here since kotlin will see the `isNullOrBlank` above and do a smart cast to `String`

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/maintain/AuthUserService.kt
```diff
@@ -245,8 +241,8 @@ open class AuthUserService(private val userRepository: UserRepository,
     validate(firstName, lastName)
     // will always be a user at this stage since we're retrieved it from the authentication
     val user = userRepository.findByUsernameAndSource(username, AuthSource.auth).orElseThrow()
-    user.person.firstName = firstName
-    user.person.lastName = lastName
+    user.person.firstName = firstName?.trim()
```
petergphillips: although the smart cast doesn't work here since it can't tell that `validate` will have thrown the exception if `firstName` is null

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/resource/UserDetailsController.kt
```diff
@@ -40,12 +40,6 @@ open class UserDetailsController(private val authUserService: AuthUserService,
                          authentication: Authentication,
                          request: HttpServletRequest, response: HttpServletResponse): ModelAndView {
 
-    val modelAndView = ModelAndView("userDetails").validateBlankInputs(firstName, lastName)
```
petergphillips: validation moved to service

src/main/resources/templates/fragments/header.html
```diff
@@ -28,7 +28,7 @@
         <nav>
           <ul id="navigation" class="govuk-header__navigation " aria-label="Top Level Navigation">
             <li class="govuk-header__navigation-item govuk-header__navigation-item--active">
-              <span sec:authentication="principal.name" id="principal-name">Name</span>
+              <span th:text="${#authentication?.principal?.name}" id="principal-name">Name</span>
```
petergphillips: `sec:authentication` doesn't escape the name so have to use thymeleaf instead

src/main/resources/templates/userDetails.html
```diff
@@ -38,7 +38,7 @@ <h1 class="govuk-heading-xl">Update personal details</h1>
         </span>
         <input th:class="'govuk-input' + (${error_firstName} ? ' govuk-input--error' : '')" id="firstName"
                name="firstName"
-               type="text" data-element-id="firstName" autofocus autocomplete="firstName" spellcheck="false"
+               type="text" data-element-id="firstName" autofocus autocomplete="given-name" spellcheck="false"
```
petergphillips: recommended autocomplete values from https://design-system.service.gov.uk/patterns/names/

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/maintain/AuthUserService.kt
mikehalmamoj: IsNullOrBlank triggers a smart cast?  Nice.

PR #353, opened 2020-01-31
DT-517: Add in expired password functionality

petergphillips:


build.gradle
```diff
@@ -188,6 +188,7 @@ drivers.each { driver ->
 
 test {
   exclude '**/integration/*'
+  useJUnitPlatform()
```
petergphillips: for junit 5

src/test/java/uk/gov/justice/digital/hmpps/oauth2server/resource/user/ExistingPasswordControllerTest.kt
```diff
@@ -0,0 +1,114 @@
...
+
+class ExistingPasswordControllerTest {
+  private val authenticationManager: AuthenticationManager = mock()
+  private val tokenService: TokenService = mock()
+  private val telemetryClient: TelemetryClient = mock()
+  private val controller = ExistingPasswordController(authenticationManager, tokenService, telemetryClient)
+  private val token = TestingAuthenticationToken(UserDetailsImpl("user", "name", setOf(), AuthSource.auth.name, null), "pass")
+
+  @Nested
+  inner class ExistingPasswordRequest {
```
petergphillips: quite like this separation in junit 5 - create `Nested` per method being tested

src/main/resources/templates/changePassword.html
```diff
@@ -43,15 +43,20 @@ <h2 class="govuk-error-summary__title" id="error-summary">
...
   </p>
 
   <form class="form" th:action="@{/change-password}" method="POST" id="changePasswordForm">
-    <div class="govuk-form-group">
+    <div class="govuk-form-group govuk-visually-hidden">
       <input id="token" name="token" type="hidden" data-element-id="token"
              th:value="${#strings.isEmpty(token) ? param.token : token}">
+      <input id="existing" type="hidden" data-element-id="existing" th:value="${expired}"/>
+
+      <label class="govuk-label" for="username">Username</label>
```
petergphillips: these are needed so that password managers know what password is being changed.  We were getting a chrome error on this page telling us we were missing the username.  Not used by the controller, but will add in next commit.

src/main/resources/templates/user/existingPassword.html
```diff
@@ -0,0 +1,76 @@
...
+  </div>
+
+
+  <h1 class="govuk-heading-xl">Change your password</h1>
+
+  <p class="govuk-body">Confirm your current password to continue.</p>
+
+  <form th:action="@{/existing-password}" method="POST" id="existingpasswordForm" class="govuk-input-w">
+    <div class="govuk-form-group govuk-visually-hidden">
+      <label class="govuk-label" for="username">Username</label>
```
petergphillips: again needed so that password managers know what username the password is for.  Tried `hidden` but didn't work, so settled for `text` with a govuk style around it (hence added label for screenreaders)

build.gradle
andymarke: Is that the only change required to use junit 5 (other than jar dependencies)?

petergphillips: yes, spring boot these days adds everything else in for us

PR #363, opened 2020-02-07
DT-544 tests to kotlin

petergphillips:


src/test/java/uk/gov/justice/digital/hmpps/oauth2server/config/AuthenticationManagerConfigurationTest.kt
```diff
@@ -0,0 +1,47 @@
...
+  private val userStateAuthenticationFailureHandle: UserStateAuthenticationFailureHandler = mock()
+  private val authenticationManagerBuilder: AuthenticationManagerBuilder = mock()
+  private var authenticationManagerConfiguration = AuthenticationManagerConfiguration(nomisUserDetailsService, authUserDetailsService,
+      deliusUserDetailsService, accessDeniedHandler, logoutSuccessHandler, jwtAuthenticationSuccessHandler, jwtCookieAuthenticationFilter,
+      jwtCookieName, cookieRequestCache, authAuthenticationProvider, nomisAuthenticationProvider, deliusAuthenticationProvider,
+      userStateAuthenticationFailureHandle)
+
+  @Test
+  fun configure_deliusProviderIsLast() {
+    argumentCaptor<LockingAuthenticationProvider>().apply {
```
andymarke: Not convinced this makes the code  more comprehensible - feels like a neat trick. Mostly because the Captor isn't the main subject of this method - I had to really think to understand what was going on.
Maybe see what other people think

src/test/java/uk/gov/justice/digital/hmpps/oauth2server/verify/VerifyEmailServiceTest.kt
```diff
@@ -70,7 +67,6 @@ class VerifyEmailServiceTest {
   }
 
   @Test
-  @Throws(NotificationClientException::class, VerifyEmailException::class)
```
andymarke: How is this assert done if not via this annotation?

petergphillips: this isn't on the `@Test` annotation.  Kotlin doesn't have checked exceptions so this is a hint to java land - see https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.jvm/-throws/index.html and therefore can be ignored in a test

src/test/java/uk/gov/justice/digital/hmpps/oauth2server/config/AuthenticationManagerConfigurationTest.kt
petergphillips: I agree, wasn't a great fan will revert in next PR

src/test/java/uk/gov/justice/digital/hmpps/oauth2server/verify/VerifyEmailServiceTest.kt
andymarke: Ah, yes - makes senses. An agreed makes no sense for a test and presumbly would be used for APIs written in Kotlin being used by Java.

PR #365, opened 2020-02-17
DT-515: added user self service change email address

SimonMitchellMOJ:


build.gradle
```diff
@@ -125,11 +125,11 @@ dependencies {
   implementation 'nz.net.ultraq.thymeleaf:thymeleaf-layout-dialect:2.4.1'
   implementation 'uk.gov.service.notify:notifications-java-client:3.15.1-RELEASE'
 
-  implementation 'org.flywaydb:flyway-core:6.2.1'
+  implementation 'org.flywaydb:flyway-core:6.2.3'
   implementation 'com.zaxxer:HikariCP:3.4.2'
   implementation 'org.apache.commons:commons-lang3:3.9'
   implementation 'org.apache.commons:commons-text:1.8'
-  implementation 'com.microsoft.sqlserver:mssql-jdbc:8.2.0.jre11'
+  implementation 'com.microsoft.sqlserver:mssql-jdbc:8.2.0.jre13'
```
brightonsbox: Are we definitely using JRE13 yet?

SimonMitchellMOJ: I will revert, I didn't mean to update that dependency

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/resource/ExistingPasswordController.kt
```diff
@@ -24,45 +24,51 @@ class ExistingPasswordController(private val authenticationManager: Authenticati
...
   }
 
-  private fun continueToNewPassword(username: String): ModelAndView {
+  private fun continueToNewEmailOrPassword(username: String, type: String): ModelAndView {
     // successfully logged in with credentials, so generate change password token
     val token = tokenService.createToken(TokenType.CHANGE, username)
 
-    // and take them to existing change passsword pages to continue flow
-    return ModelAndView("redirect:/new-password", "token", token)
+    return ModelAndView("redirect:/new-$type", "token", token)
```
brightonsbox: Potentially worth validating the type immediately when received at the API otherwise it looks like a bad type would cause a 404 being thrown in these cases?

brightonsbox: As discussed, probably doesn't matter since this is only expected to be called by other parts of this same codebase

PR #373, opened 2020-02-28
DT-516: self-serve mobile user re-enters verified mobile number, doesn…

SimonMitchellMOJ:
…'t save/send verify code but takes user to Already verified page


src/main/java/uk/gov/justice/digital/hmpps/oauth2server/security/UserService.kt
```diff
@@ -40,4 +40,12 @@ open class UserService(private val nomisUserService: NomisUserService,
     val user: User = findUser(userDetails.username).orElseGet { userDetails.toUser() }
     return StringUtils.isNotEmpty(user.email) && user.isVerified
   }
+
+  open fun isSameAsCurrentVerifiedMobile(username: String, mobile: String): Boolean {
+    val user: User = findUser(username).orElseThrow()
+    val canonicalMobile = mobile.replace("\\s+".toRegex(), "")
+    return if (!user.isMobileVerified) {
+      false
+    } else canonicalMobile == user.mobile
```
petergphillips: return
```kotlin
return user.isMobileVerified && canonicalMobile == user.mobile
```

PR #379, opened 2020-03-03
DT-370 Send the kid header back on generated Oauth tokens

mikehalmamoj:


src/test/groovy/uk/gov/justice/digital/hmpps/oauth2server/api/specs/OauthSpecification.groovy
```diff
@@ -353,4 +354,16 @@ class OauthSpecification extends TestSpecification {
...
+    def "Kid header is returned"() {
+
+        given:
+        def oauthRestTemplate = getOauthPasswordGrant("ITAG_USER", "password", "elite2apiclient", "clientsecret")
+
+        when:
+        def accessToken = oauthRestTemplate.getAccessToken()
+
+        then:
+        new JWTDecoder(accessToken.value).getHeaderClaim("kid") != null
```
petergphillips: could you assert the value too - isn't it set in application.yml?

PR #382, opened 2020-03-05
DT-524: Move more tests over to fluentlenium and separate build proce…

petergphillips:
…sses


.circleci/config.yml
```diff
@@ -1,7 +1,7 @@
 defaults: &defaults
   working_directory: ~/app
   environment:
-    JAVA_TOOL_OPTIONS: -Xmx256m -XX:ParallelGCThreads=2 -XX:ConcGCThreads=2 -XX:ParallelGCThreads=2 -Djava.util.concurrent.ForkJoinPool.common.parallelism=2 -Dorg.gradle.daemon=false -Dorg.gradle.jvmargs=-XX:+UseContainerSupport
+    _JAVA_OPTIONS: -Xmx256m -XX:ParallelGCThreads=2 -XX:ConcGCThreads=2 -XX:ParallelGCThreads=2 -Djava.util.concurrent.ForkJoinPool.common.parallelism=2 -Dorg.gradle.daemon=false -Dorg.gradle.jvmargs=-XX:+UseContainerSupport
```
petergphillips: this is oracle specific, but overrides all other args supplied on command line, therefore is enforced.

.circleci/config.yml
```diff
@@ -67,6 +67,9 @@ jobs:
       - run:
           name: Run Integration Tests
           command: ./gradlew testIntegration
+      - run:
+          name: Run Fluent Integration Tests
+          command: ./gradlew testFluentIntegration
```
petergphillips: separate out to try to reduce memory usage of chrome too

PR #384, opened 2020-03-06
DT-524: Gradually start correcting account details flow

petergphillips:


src/main/resources/templates/account/accountDetails.html
```diff
@@ -31,7 +31,7 @@ <h1 class="govuk-heading-xl">Account details</h1>
         Sarah Philips
       </dd>
       <dd class="govuk-summary-list__actions">
-        <a th:href="@{/change-name}" th:if="${authUser.authSource == 'auth'}" data-qa="changeName">
+        <a th:href="@{/user-details}" th:if="${authUser.authSource == 'auth'}" data-qa="changeName">
```
petergphillips: this should be change name, but will do so in a future PR

src/main/resources/templates/account/accountDetails.html
```diff
@@ -50,15 +50,14 @@ <h1 class="govuk-heading-xl">Account details</h1>
...
     <div class="govuk-summary-list__row">
       <dt class="govuk-summary-list__key">
-        Password expiry
+        Password
       </dt>
-      <dd class="govuk-summary-list__value" data-qa="passwordExpiry"
-          th:text="${#temporals.format(authUser.passwordExpiry, 'd MMMM yyyy HH:mm')}">
-        5 August 2039 - 20:52:13
+      <dd class="govuk-summary-list__value" data-qa="passwordExpiry">
+        ************
```
petergphillips: password expiry doesn't mean anything for auth users and we don't get the information for delius users.  Leaving out for now until there is a requirement...

src/main/resources/templates/account/accountDetails.html
```diff
@@ -129,7 +128,7 @@ <h2 class="govuk-heading-m" id="authentication">Authentication</h2>
   </dl>
 
   <div class="govuk-form-group">
-    <a class="govuk-button" role="button" data-qa="cancel" th:href="@{/account-details-cancel}">Cancel</a>
+    <a class="govuk-button" role="button" id="cancel" data-qa="cancel" th:href="@{/account-details-cancel}">Cancel</a>
```
petergphillips: need this to get the styling correct so it isn't a green button

src/test/java/uk/gov/justice/digital/hmpps/oauth2server/integration/AbstractAuthSpecification.kt
```diff
@@ -53,7 +53,7 @@ open class AbstractAuthSpecification : FluentTest() {
   }
 }
 
-open class AuthPage(val title: String, val heading: String) : FluentPage() {
+open class AuthPage<T>(val title: String, val heading: String) : FluentPage() {
```
petergphillips: fluentlenium isn't very fluent without being able to return `this` e.g. at present `isAt` returns `void` which means calls then can't be chained.  By parameterising with the page class we can then return `this as T` and start the chaining

src/test/java/uk/gov/justice/digital/hmpps/oauth2server/integration/AccountDetailsSpecification.kt
```diff
@@ -43,16 +43,28 @@ class AccountDetailsSpecification : AbstractAuthSpecification() {
...
 }
 
 @PageUrl("/account-details")
-class AccountDetailsPage : AuthPage("HMPPS Digital Services - Account Details", "Account details") {
+class AccountDetailsPage : AuthPage<AccountDetailsPage>("HMPPS Digital Services - Account Details", "Account details") {
+  @Suppress("UsePropertyAccessSyntax")
   fun checkAuthDetails(): AccountDetailsPage {
     assertThat(el("[data-qa='username']").text()).isEqualTo("AUTH_RO_USER")
     assertThat(el("[data-qa='name']").text()).isEqualTo("Ryan-Auth Orton")
     assertThat(el("[data-qa='lastLoggedIn']").text()).isNotBlank()
```
petergphillips: intellij tells me to use property access syntax here even though `isNotBlank` is actually a function.  Changing to property access then causes the class not to compile :man_facepalming:

PR #385, opened 2020-03-06
DT-524: Switch to back link instead of cancel for account details

petergphillips:


src/main/java/uk/gov/justice/digital/hmpps/oauth2server/resource/ExistingPasswordController.kt
```diff
@@ -36,7 +35,7 @@ class ExistingPasswordController(private val authenticationManager: Authenticati
         .addObject("error", "required")
         .addObject("type", type)
 
-    val username = getUserName(authentication)
+    val username = authentication.name
```
petergphillips: can just call `name` on `authentication` instead :grinning:

src/test/java/uk/gov/justice/digital/hmpps/oauth2server/integration/AccountDetailsSpecification.kt
```diff
@@ -12,7 +12,7 @@ class AccountDetailsSpecification : AbstractAuthSpecification() {
 
   @Test
   fun `auth account details`() {
-    goTo(loginPage).loginAs("AUTH_RO_USER", "password123456")
+    goTo(loginPage).loginAs("AUTH_RO_USER")
```
petergphillips: don't you just love kotlin default parameters :joy:

src/main/resources/templates/fragments/layout.html
```diff
@@ -62,7 +62,7 @@
 <th:block th:replace="fragments/header :: header"/>
 
 <div class="govuk-width-container">
-  <a th:if="${showBack}" th:href="@{/login}" class="govuk-back-link" data-qa="back-link">Back</a>
+  <a th:if="${showBack}" th:href="@{/}" class="govuk-back-link" data-qa="back-link">Back</a>
```
petergphillips: not sure why this was going to `login`, but if they're not logged in then this will take them to the login page anyway

PR #386, opened 2020-03-06
DT-524: Switch to kotlin build file

petergphillips:


build.gradle.kts
```diff
@@ -0,0 +1,197 @@
...
+
+  val testFluentIntegration by registering(Test::class) {
+    systemProperty("fluentlenium.capabilities", """{"chromeOptions": {"args": ["headless","disable-gpu","disable-extensions","no-sandbox","disable-application-cache"]}}""")
+    useJUnitPlatform()
+    include("uk/gov/justice/digital/hmpps/oauth2server/integration/*")
+    setMaxHeapSize("256m")
+  }
+
+  val testIntegration by registering(Test::class) {
+    systemProperty("geb.env", "chromeHeadless")
```
petergphillips: I think it actually turns out that we weren't running the groovy tests in headless mode :woman_facepalming: .  Could explain why they were failing quite often

src/test/resources/GebConfig.groovy
```diff
@@ -15,12 +15,7 @@ environments {
     chromeHeadless {
         driver = {
             ChromeOptions options = new ChromeOptions()
-            options.addArguments('headless')
-            options.add_argument('disable-extensions')
```
petergphillips: these didn't compile, so can't have actually been running

build.gradle.kts
andymarke: surprised it works at all without headless mode

PR #390, opened 2020-03-10
DT-524: Move verify email specification test to fluentlenium

petergphillips:


src/main/java/uk/gov/justice/digital/hmpps/oauth2server/resource/account/ChangeMobileController.kt
```diff
@@ -30,7 +30,7 @@ class ChangeMobileController(private val userService: UserService,
   fun changeMobile(@RequestParam mobile: String?, authentication: Authentication): ModelAndView {
     val username = authentication.name
     if (userService.isSameAsCurrentVerifiedMobile(username, mobile)) {
-      return ModelAndView("verifyMobileAlready")
+      return ModelAndView("redirect:/verify-mobile-already")
```
petergphillips: need a redirect so that page url can be tested in fluentlenium - otherwise page can have multiple urls depending on where come from

src/test/java/uk/gov/justice/digital/hmpps/oauth2server/integration/VerifyMobileSpecification.kt
```diff
@@ -0,0 +1,97 @@
...
+  @Test
+  fun `Change mobile flow invalid verification code`() {
+
+    goTo(loginPage).loginAs("AUTH_CHANGE_MOBILE_UPDATE")
+
+    goTo(changeMobilePage)
+        .updateMobileAs("07700900322", "07700900321")
+
+    verifyMobileSentPage.isAtPage()
+        .submitCode("123456")
```
mikehalmamoj: Will this test fail 1 time in every 999999?

petergphillips: yes :grinning:

PR #391, opened 2020-03-10
Fix script name in docs and add a bit more detail

Nimphal:


user-load.sh
```diff
@@ -24,6 +26,8 @@
 #        4) Last name
 #        5) group code
 #
+# Make sure there is a new line at the end of the file, otherwise it will not load.
```
petergphillips: Surely this should read:
>Make sure you use a decent text editor like vim otherwise you won't get a new line at the end of the file

Mjwillis: Lol

PR #399, opened 2020-03-12
DT 634 resend mfa security code preference - Text

SimonMitchellMOJ:


src/main/java/uk/gov/justice/digital/hmpps/oauth2server/resource/MfaController.kt
```diff
@@ -79,33 +78,77 @@ open class MfaController(private val jwtAuthenticationSuccessHandler: JwtAuthent
...
+  }
+
+  @GetMapping("/mfa-resend-text")
+  fun mfaResendTextRequest(@RequestParam token: String, @RequestParam mfaPreference: MfaPreferenceType): ModelAndView {
+
+    val optionalError = tokenService.checkToken(TokenType.MFA, token)
+
+    val user = tokenService.getUserFromToken(TokenType.MFA, token)
+    val partialMobile = user.mobile.replaceRange(0, 7, "*******")
+    val partialEmail = user.email.take(6) + "******@******" + user.email.takeLast(7)
```
petergphillips: can't see any tests for the partial email / text scenarios?  Was interested to see what would happen if `mobile` is null, only contains 2 characters etc...

PR #403, opened 2020-03-16
DT-632: Add more int tests around mfa specification

petergphillips:


src/test/java/uk/gov/justice/digital/hmpps/oauth2server/integration/AbstractAuthSpecification.kt
```diff
@@ -61,11 +61,15 @@ open class AuthPage<T>(val title: String, val heading: String) : FluentPage() {
   @FindBy(css = "#error-detail")
   private lateinit var errorDetail: FluentWebElement
 
-  internal fun isAtPage(): T {
+  override fun isAt() {
```
petergphillips: no idea how I managed to remove this method :man_facepalming:

src/test/java/uk/gov/justice/digital/hmpps/oauth2server/integration/MfaSpecification.kt
```diff
@@ -99,7 +117,7 @@ class MfaSpecification : AbstractAuthSpecification() {
   @Test
   fun `MFA code is incorrect text`() {
     goTo(loginPage)
-        .loginWithMfaEmail("AUTH_MFA_PREF_TEXT2")
+        .loginWithMfaText("AUTH_MFA_PREF_TEXT2")
```
petergphillips: yes these were all working previously because I accidentally removed the overridden `isAt` method

PR #404, opened 2020-03-16
Dt 634 resend mfa security code

SimonMitchellMOJ:


src/main/java/uk/gov/justice/digital/hmpps/oauth2server/service/MfaService.kt
```diff
@@ -115,6 +116,30 @@ class MfaService(@Value("\${application.authentication.mfa.whitelist}") whitelis
     return code
   }
 
+  fun buildModelAndViewWithMfaResendOptions(token: String, mfaPreference: MfaPreferenceType): ModelAndView {
+    val user = tokenService.getUserFromToken(TokenType.MFA, token)
+    val modelAndView = ModelAndView("mfaResend", "token", token)
+        .addObject("mfaPreference", mfaPreference)
+
+    if (user.isVerified) {
```
petergphillips: why you checking `verified` here?  This is different from the html where you're checking that the field isn't empty.
Might be better to just remove both checks and let the front end display if they are set

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/service/MfaService.kt
```diff
@@ -115,6 +116,30 @@ class MfaService(@Value("\${application.authentication.mfa.whitelist}") whitelis
...
+      modelAndView.addObject("mobile", getMaskedMobile(user))
+    }
+    return modelAndView
+  }
+
+  fun getMaskedMobile(user: User): String {
+    return user.mobile.replaceRange(0, 7, "*******")
+  }
+
+  fun getMaskedEmail(user: User): String {
```
petergphillips: these would be better on the `User` object instead

PR #413, opened 2020-03-19
DT-660: sqlserver specific removal of fields

petergphillips:


src/main/java/uk/gov/justice/digital/hmpps/oauth2server/config/FlywayConfig.kt
```diff
@@ -0,0 +1,66 @@
...
+                 @Value("\${auth.flyway.locations}") flywayLocations: List<String>): Flyway {
+    val flyway = Flyway.configure()
+        .dataSource(authDataSource)
+        .locations(*replaceVendorLocations(flywayLocations, authDataSource).toTypedArray())
+        .load()
+    flyway.migrate()
+    return flyway
+  }
+
+  private fun replaceVendorLocations(locations: List<String>, dataSource: DataSource): List<String> {
```
petergphillips: code nabbed from `org.springframework.boot.autoconfigure.flyway.FlywayAutoConfiguration`, which we can't use since we have multiple data sources

src/main/resources/db/auth_sqlserver/V3_20__user_remove_mobile_fields.sql
```diff
@@ -0,0 +1,15 @@
+DECLARE @ConstraintName nvarchar(200);
+
+SELECT @ConstraintName = Name FROM SYS.DEFAULT_CONSTRAINTS
+WHERE PARENT_OBJECT_ID = OBJECT_ID('users')
+AND PARENT_COLUMN_ID = (SELECT column_id FROM sys.columns
+    WHERE NAME = N'mobile_verified'
+    AND object_id = OBJECT_ID(N'users'));
+
+IF @ConstraintName IS NOT NULL
+    EXEC('ALTER TABLE users DROP CONSTRAINT ' + @ConstraintName)
```
petergphillips: have to drop the `default 0` on the mobile_verified field before can drop the field.  Constraint was named dynamically so has different names in all environments

PR #414, opened 2020-03-19
WIP DT-635 Attempt to find the user even for proxy-users.

mikehalmamoj:
If we know that the token has been issued to a known user then any
consuming service can decide how much to trust that user.  For example
we might want to let Auth users perform certain updates that we'd reject
for an unknown proxy-user.


mikehalmamoj:
Sorry this was meant to be a draft PR (hence the WIP - but that's a GitLab thing).  Please do not merge!!!

PR #415, opened 2020-03-20
DT-630: Move some int tests to kotlin

petergphillips:


src/test/java/uk/gov/justice/digital/hmpps/oauth2server/resource/api/AuthUserIntTest.kt
```diff
@@ -0,0 +1,150 @@
...
+        .get().uri("/auth/api/authuser/search?name=test2")
+        .headers(setAuthorisation("AUTH_ADM"))
+        .exchange()
+        .expectStatus().isOk
+        .expectHeader().contentType(MediaType.APPLICATION_JSON)
+        .expectBody()
+        .json("auth_user_search.json".readFile())
+  }
+
+  private fun String.readFile(): String = this@AuthUserIntTest::class.java.getResource(this).readText()
```
petergphillips: note the strange `this` annotation.  `this` in scope is the String since we've created an extension function

src/test/java/uk/gov/justice/digital/hmpps/oauth2server/resource/api/AuthUserIntTest.kt
```diff
@@ -0,0 +1,150 @@
...
+
+    webTestClient
+        .get().uri("/auth/api/user/$username")
+        .headers(setAuthorisation("ITAG_USER_ADM"))
+        .exchange()
+        .expectStatus().isOk
+        .expectHeader().contentType(MediaType.APPLICATION_JSON)
+        .expectBody(object : ParameterizedTypeReference<Map<String, String>>() {})
+        .consumeWith<Nothing?> {
+          assertThat(it!!.responseBody!!.filter { it.key != "userId" }).containsExactlyEntriesOf(
```
mikehalmamoj: That's not very pretty - better with a JSONPath?

petergphillips: agreed, I was playing around with it to see what it will look like, will change on next PR

PR #420, opened 2020-03-23
DT-665 Add sub to attributes for JWT.

Mjwillis:


src/test/java/uk/gov/justice/digital/hmpps/oauth2server/config/JWTTokenEnhancerTest.kt
```diff
@@ -17,31 +17,33 @@ import uk.gov.justice.digital.hmpps.oauth2server.security.UserDetailsImpl
 import java.util.*
 
 internal class JWTTokenEnhancerTest {
-  private val authentication: OAuth2Authentication = mock()
+    private val authentication: OAuth2Authentication = mock()
```
petergphillips: your kotlin formatter is incorrect - should be 2 chars for kotlin, not 4

PR #427, opened 2020-03-31
DT-688: added secondary email as mfa preference

SimonMitchellMOJ:


src/main/java/uk/gov/justice/digital/hmpps/oauth2server/auth/model/User.java
```diff
@@ -195,8 +195,14 @@ public String getMaskedEmail() {
         return email.substring(0, emailCharactersReduced) + "******@******" + email.substring(email.length() - 7);
     }
 
+    public String getMaskedSecondaryEmail() {
```
petergphillips, line 198: would prefer a masked email function rather than duplicating the code block

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/auth/model/User.java
```diff
@@ -395,6 +405,9 @@ public boolean hasVerifiedMfaMethod() {
         findContact(ContactType.MOBILE_PHONE)
                 .filter(c -> StringUtils.isNotBlank(c.getValue()) && c.getVerified())
                 .ifPresent(c -> preferences.add(MfaPreferenceType.TEXT));
+        findContact(ContactType.SECONDARY_EMAIL)
+                .filter(c -> StringUtils.isNotBlank(c.getValue()) && c.getVerified())
```
petergphillips, line 409: just out of interest - do you think we need the blank check?

PR #429, opened 2020-04-01
DT-701: Fix delius users not being able to get refresh tokens

petergphillips:


src/main/java/uk/gov/justice/digital/hmpps/oauth2server/security/DeliusUserDetailsService.kt
```diff
@@ -1,22 +1,19 @@
...
 import org.springframework.security.core.userdetails.UsernameNotFoundException
 import org.springframework.security.web.authentication.preauth.PreAuthenticatedAuthenticationToken
 import org.springframework.stereotype.Component
 import org.springframework.stereotype.Service
-import org.springframework.transaction.annotation.Transactional
 import uk.gov.justice.digital.hmpps.oauth2server.delius.service.DeliusUserService
 import uk.gov.justice.digital.hmpps.oauth2server.service.MfaService
 
 @Service("deliusUserDetailsService")
-@Transactional(readOnly = true)
```
petergphillips, line 17: don't need a transaction as going to community api

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/security/NomisUserDetailsService.kt
```diff
@@ -15,8 +15,8 @@ import javax.persistence.EntityManager
 import javax.persistence.PersistenceContext
 
 @Service("nomisUserDetailsService")
-@Transactional(readOnly = true)
-open class NomisUserDetailsService(private val nomisUserService: NomisUserService) :
+@Transactional(readOnly = true, noRollbackFor = [UsernameNotFoundException::class])
```
petergphillips, line 18: user not found here causing rollback marker to be set here so subsequent delius call was then attempted to be rolled back

PR #432, opened 2020-04-03
DT 621 change initial password email changes

SimonMitchellMOJ:
update firstName parameter for change initial password email to take full name 
and added fullName to notify parameters for all email parameter sets which have firstName in already

### DIFF
--- a/src/main/java/uk/gov/justice/digital/hmpps/oauth2server/maintain/AuthUserService.kt
+++ b/src/main/java/uk/gov/justice/digital/hmpps/oauth2server/maintain/AuthUserService.kt
@@ -111,7 +111,7 @@ class AuthUserService(private val userRepository: UserRepository,
     val setPasswordLink = url + userToken.token
     val username = user.username
     val email = user.email
-    val parameters = mapOf("firstName" to user.firstName, "fullName" to user.name, "resetLink" to setPasswordLink, "supportLink" to supportLink)
+    val parameters = mapOf("firstName" to user.name, "fullName" to user.name, "resetLink" to setPasswordLink, "supportLink" to supportLink)
     // send the email
     try {
       log.info("Sending initial set password to notify for user {}", username)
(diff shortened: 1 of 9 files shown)


src/main/java/uk/gov/justice/digital/hmpps/oauth2server/maintain/AuthUserService.kt
```diff
@@ -111,7 +111,7 @@ class AuthUserService(private val userRepository: UserRepository,
     val setPasswordLink = url + userToken.token
     val username = user.username
     val email = user.email
-    val parameters = mapOf("firstName" to user.firstName, "fullName" to user.name, "resetLink" to setPasswordLink, "supportLink" to supportLink)
+    val parameters = mapOf("firstName" to user.name, "fullName" to user.name, "resetLink" to setPasswordLink, "supportLink" to supportLink)
```
andymarke, line 114: not sure if I understand why this map contains "fullName" and "firstName" both set to user.name

petergphillips: because the existing templates use `firstName` and we can't change them until we have gone to production with this change

petergphillips: and long term having the full name in a `firstName` field will just be confusing

SimonMitchellMOJ: Currently in Notify we have a place holder "Dear ((firstName))" we are changing this to be fullName but because the template is current being used in production I cannot just update the template so I have cheated and just altered the property value to see what it is like in dev before we move it though to prod and then once it is there I will alter the notify template so that the placeholder/parameter/parameter value all match

PR #444, opened 2020-04-20
DT-717: Add in delius health and prepare for multiple wiremocks

petergphillips:


src/main/java/uk/gov/justice/digital/hmpps/oauth2server/config/AuthenticationManagerConfiguration.java
```diff
@@ -128,7 +128,7 @@ public void configure(final WebSecurity web) {
         web
                 .ignoring()
                 .antMatchers("/css/**", "/js/**", "/images/**", "/fonts/**", "/webjars/**", "/favicon.ico",
-                        "/health", "/health/ping", "/info", "/ping", "/error", "/terms", "/contact", "/change-password",
+                        "/health/**", "/info", "/ping", "/error", "/terms", "/contact", "/change-password",
```
petergphillips, line 131: allow calls to health and everything underneath

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/service/HealthCheck.kt
```diff
@@ -0,0 +1,31 @@
...
+}
+
+@Component
+class DeliusApiHealth(@Qualifier("deliusApiHealthRestTemplate") private val restTemplate: RestTemplate) : HealthCheck(restTemplate) {
+  override fun health(): Health {
+    val health = super.health()
+    if (health.status != Status.DOWN) return health
+
+    // can still run a degraded service with delius down, so mark as healthy with error message
+    return Health.up().withDetails(health.details).build()
```
mikehalmamoj, line 29: As this is a little unusual I wonder if we should have a test for it?  In case someone ignores the comment and "fixes" the health status.  (But I guess they'd just "fix" the test too, but at least that would require a conscious effort to ignore)

petergphillips: do you mean src/test/java/uk/gov/justice/digital/hmpps/oauth2server/resource/HealthDeliusDownIntTest.kt (see lower down)

src/test/java/uk/gov/justice/digital/hmpps/oauth2server/resource/HealthDeliusDownIntTest.kt
```diff
@@ -0,0 +1,16 @@
...
+
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.jupiter.api.Test
+
+class HealthDeliusDownIntTest : IntegrationTest() {
+  @Test
+  fun `Health reports delius info`() {
+    webTestClient.get().uri("/auth/health/deliusApiHealth")
+        .exchange()
+        .expectBody().jsonPath("status").isEqualTo("UP")
```
petergphillips, line 11: @mikehalmamoj here - although should really have 
```java
        .expectStatus().isOk
```
line as well, will add in next PR

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/service/HealthCheck.kt
mikehalmamoj: Yeah, sorry I missed that.

PR #455, opened 2020-04-27
DT-773: Upgrade spring security and schedule deletion of old auth codes

petergphillips:


src/main/java/uk/gov/justice/digital/hmpps/oauth2server/timed/BatchUserProcessor.java
```diff
@@ -25,7 +25,7 @@ BatchState findAndProcessInBatches() {
                 batch.resetProcessedAndIncrementTotal(service.processInBatches());
             } catch (final Exception e) {
                 // have to catch the exception here otherwise scheduling will stop
-                log.error("Caught exception {} for {}", e.getClass().getSimpleName(), telemetryClient, e);
+                log.error("Caught exception {} for {}", e.getClass().getSimpleName(), telemetryPrefix, e);
```
petergphillips, line 28: typo on log message

src/main/resources/db/auth/V4_00__oauth_code_recreate.sql
```diff
@@ -0,0 +1,9 @@
+DROP TABLE IF EXISTS oauth_code;
+
+create table oauth_code
+(
+    code           VARCHAR(256) NOT NULL PRIMARY KEY,
+    authentication VARBINARY( MAX),
+    created_date   DATETIME2    NOT NULL DEFAULT CURRENT_TIMESTAMP
```
petergphillips, line 7: old table had `created_date` as a row version column by accident so easiest to remove and recreate to get the correct column and trash all the old data

src/main/resources/db/dev/data/auth/V900_0__oauth_sample_data.sql
```diff
@@ -282,4 +282,4 @@ INSERT INTO user_contact (user_id, type, value, verified) SELECT user_id, 'MOBIL
...
...001171007e001371007e001571007e00177871007e003a707073720042756b2e676f762e6a7573746963652e6469676974616c2e686d7070732e6f61757468327365727665722e73656375726974792e5573657244657461696c73496d706c00000000000000010200044c000a61757468536f7572636571007e000e4c00056a7774496471007e000e4c00046e616d6571007e000e4c000675736572496471007e000e787200326f72672e737072696e676672616d65776f726b2e73656375726974792e636f72652e7573657264657461696c732e5573657200000000000002120200075a00116163636f756e744e6f6e457870697265645a00106163636f756e744e6f6e4c6f636b65645a001563726564656e7469616c734e6f6e457870697265645a0007656e61626c65644c000b617574686f72697469657371007e001c4c000870617373776f726471007e000e4c0008757365726e616d6571007e000e7870010101017371007e002b737200116a6176612e7574696c2e54726565536574dd98509395ed875b0300007870737200466f72672e737072696e676672616d65776f726b2e73656375726974792e636f72652e7573657264657461696c732e5573657224417574686f72697479436f6d70617261746f720000000000000212020000787077040000000571007e000f71007e001171007e001371007e001571007e00177874000074000d495441475f555345525f41444d7400056e6f6d697374002465646334356338352d623638642d343530642d393031332d61303439343063653766646274000949746167205573657274000131);
```
petergphillips, line 285: spring security forces this to change as serial version id no longer compatible

src/test/java/uk/gov/justice/digital/hmpps/oauth2server/resource/api/OauthIntTest.kt
```diff
@@ -9,17 +9,17 @@ class OauthIntTest : IntegrationTest() {
   @Test
   fun `Existing auth code stored in database can be redeemed for access token`() {
     val authCode = "03sn0c" // from database oauth_code table
-    val clientUrl = "http://localhost:3000/login/callback" // same as row in oauth_code table
+    val clientUrl = "http://localhost:8081/login" // same as row in oauth_code table
```
petergphillips, line 12: different test data for token stored in database

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/auth/model/OauthCode.kt
```diff
@@ -0,0 +1,23 @@
...
+@Entity
+@Table(name = "OAUTH_CODE")
+data class OauthCode(
+    @Id
+    @Column(nullable = false)
+    val code: String) {
+
+  @Lob
+  @Column(nullable = false)
+  var authentication: ByteArray? = null
```
andymarke, line 19: Am I right in thinking you could make these non-nullable but then initialise them to default dummy values? Which I'm not sure is any better

petergphillips: agreed, not that happy with my mapping as is, but is only mapped for deletes, rest handled by spring.  Wanted to have one field of the `code` which is the comparator for equals etc, but with two extra fields.

PR #456, opened 2020-04-28
DT-773: spring security oauth upgrade

petergphillips:


build.gradle.kts
```diff
@@ -94,7 +94,7 @@ dependencies {
 
   implementation("org.springframework.boot:spring-boot-starter-web")
   implementation("org.springframework.security:spring-security-jwt:1.1.0.RELEASE")
-  implementation("org.springframework.security.oauth:spring-security-oauth2:2.4.0.RELEASE")
```
petergphillips, line 97: In this change they altered the JTI from a uuid to a base64 encoded secure random - https://github.com/spring-projects/spring-security-oauth/commit/ff555251805d472ca24ad4463b680b02554a7470

src/test/java/uk/gov/justice/digital/hmpps/oauth2server/integration/ClientLoginSpecification.kt
```diff
@@ -87,11 +87,12 @@ class ClientLoginSpecification : AbstractAuthSpecification() {
     clientAccess()
         .jsonPath(".['refresh_token','access_token']").value<JSONArray> {
           val accessJwtid = JWTParser.parse((it[0] as Map<*, *>)["access_token"].toString()).jwtClaimsSet.jwtid
+          val encodedAccessJwtId = accessJwtid.replace("/", "%2F")
 
           getRefreshToken((it[0] as Map<*, *>)["refresh_token"].toString())
               .jsonPath(".sub").isEqualTo("AUTH_USER")
               .jsonPath(".access_token").value<JSONArray> { accessToken ->
-                tokenVerificationApi.verify(postRequestedFor(urlPathMatching("/token/refresh/${accessJwtid}"))
+                tokenVerificationApi.verify(postRequestedFor(urlEqualTo("/token/refresh/${encodedAccessJwtId}"))
```
petergphillips, line 95: the `slash` is replaced in the request so that it isn't treated as a path separator

PR #459, opened 2020-04-30
DT-738:  amended backup email pages to match prototype and reduced al…

SimonMitchellMOJ:
…l headings from xl to l


src/test/java/uk/gov/justice/digital/hmpps/oauth2server/resource/VerifyEmailControllerTest.kt
```diff
@@ -69,12 +69,12 @@ class VerifyEmailControllerTest {
     verify(jwtAuthenticationSuccessHandler).proceed(request, response, principal)
   }
 
-  @Test
-  fun verifyEmailContinue() {
-    SecurityContextHolder.getContext().authentication = principal
-    verifyEmailController.verifyEmailContinue(request, response)
-    verify(jwtAuthenticationSuccessHandler).proceed(request, response, principal)
-  }
+//  @Test
```
petergphillips, line 72: commented out code? Just remove

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/resource/VerifyEmailController.java
```diff
@@ -79,11 +79,6 @@ public ModelAndView verifyEmailRequest(final Principal principal, final HttpServ
         return modelAndView;
     }
 
-    @GetMapping("/verify-email-continue")
```
petergphillips, line 82: why has this gone?

petergphillips: what about the primary flow when they are interrupted on login due to not verifying their email address?

SimonMitchellMOJ: that process uses a difference endpoint to skip - `@GetMapping("/verify-email-skip")`, all the `@GetMapping("/verify-email-continue")` did was proceed to the homepage

PR #460, opened 2020-04-30
DT-774: Minor tweaks to initial and reset password

petergphillips:


src/test/java/uk/gov/justice/digital/hmpps/oauth2server/integration/InitialPasswordSpecification.kt
```diff
@@ -87,7 +91,10 @@ class InitialPasswordSpecification : AbstractAuthSpecification() {
 }
 
 @PageUrl("/initial-password")
-class InitialPasswordPage : AuthPage<InitialPasswordPage>("HMPPS Digital Services - Create a password", "Create a password") {
+class InitialPasswordPage : SetPasswordPage()
+
+@PageUrl("/set-password")
+open class SetPasswordPage : AuthPage<SetPasswordPage>("HMPPS Digital Services - Create a password", "Create a password") {
```
petergphillips, line 97: they start on the initial page, but the submit for the page takes them to the set password page instead (as shared with reset password)

PR #463, opened 2020-05-05
DT-804: Fix more kotlin nulls found in logs

petergphillips:


src/main/java/uk/gov/justice/digital/hmpps/oauth2server/resource/ChangePasswordController.kt
```diff
@@ -31,21 +31,21 @@ class ChangePasswordController(private val jwtAuthenticationSuccessHandler: JwtA
...
+      else createModelWithTokenUsernameAndIsAdmin(TokenType.CHANGE, token, "changePassword")
 
   @PostMapping("/change-password", "/new-password")
-  fun changePassword(@RequestParam token: String?,
+  fun changePassword(@RequestParam token: String,
                      @RequestParam newPassword: String?, @RequestParam confirmPassword: String?,
                      request: HttpServletRequest, response: HttpServletResponse,
                      @RequestParam expired: Boolean?): ModelAndView? {
 
-    if (token.isNullOrBlank()) return ModelAndView("redirect:/")
```
petergphillips, line 47: added in last PR, should never happen as is a `POST` instead of our form, rather than user navigation.  Therefore removed.

PR #464, opened 2020-05-06
DT-806: use published gradle plugin

SimonMitchellMOJ:


build.gradle.kts
```diff
@@ -1,116 +1,40 @@
...
-  rejectVersionIf {
-    isNonStable(candidate.version) && !isNonStable(currentVersion)
-  }
-}
-
-group = "uk.gov.justice.digital.hmpps"
-
 val todaysDate: String = LocalDate.now().format(ISO_DATE)
-val today: Instant = Instant.now()
 version = if (System.getenv().contains("CI")) "${todaysDate}.${System.getenv("CIRCLE_BUILD_NUM")}" else todaysDate
```
mikehalmamoj, line 15: That's interesting - we might have to make that env var name configurable.

dps-gradle-spring-boot-suppressions.xml
```diff
@@ -0,0 +1,61 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
petergphillips, line 1: need to delete `dependency-check-suppress-spring.xml` as not needed any more

PR #468, opened 2020-05-11
DT-806: More hacking to fix version number

petergphillips:


.circleci/config.yml
```diff
@@ -24,6 +26,7 @@ jobs:
     executor: builder
     steps:
       - checkout
+      - hmpps/create_app_version
```
petergphillips, line 29: would switch to using hmpps orb, but still using docker hub so just grab the version stuff for now

andymarke: Nice, we could use this technique on other projects that haven't made the full switch yet

PR #470, opened 2020-05-14
DT-783: Reset password for Nomis user who has never logged in

SimonMitchellMOJ:


src/test/java/uk/gov/justice/digital/hmpps/oauth2server/verify/ResetPasswordServiceTest.kt
```diff
@@ -376,4 +383,81 @@ class ResetPasswordServiceTest {
...
   }
+
+  @Test
+  fun `Nomis User who has not logged into auth can reset password`() {
+    whenever(userRepository.findByUsername(anyString())).thenReturn(Optional.empty())
+    whenever(userService.findMasterUserPersonDetails(anyString())).thenReturn(Optional.of(buildStandardUser("user")))
+    val emailAddresses = mutableListOf("Bob.smith@hmps.gsi.gov.uk", "Bob.smith@justice.gov.uk")
+    whenever(jdbcTemplate.queryForList(anyString(), any<Class<String>>(), anyString())).thenReturn(emailAddresses)
+    val optionalLink = resetPasswordService.requestResetPassword("user", "url")
+    assertThat(optionalLink).isPresent
```
petergphillips, line 394: think you need the brackets here otherwise doesn't test anything

petergphillips: i.e. intellij is incorrect

src/test/java/uk/gov/justice/digital/hmpps/oauth2server/integration/InitialPasswordSpecification.kt
```diff
@@ -105,10 +105,31 @@ open class SetPasswordPage : AuthPage<SetPasswordPage>("HMPPS Digital Services -
   @FindBy(css = "input[type='submit']")
   private lateinit var savePassword: FluentWebElement
 
-  fun inputAndConfirmNewPassword(password: String) {
+  fun inputAndConfirmNewPassword(password: String, confirmPassword: String): SetPasswordPage {
```
petergphillips, line 108: I think you can write
```
confirmPassword: String = password
```
and it will default to the first field, meaning that you don't need to pass it in if it is the same as the `password`

src/test/java/uk/gov/justice/digital/hmpps/oauth2server/verify/ResetPasswordServiceTest.kt
```diff
@@ -376,4 +383,81 @@ class ResetPasswordServiceTest {
...
+        .staff(staff)
+        .roles(listOf(UserCaseloadRole.builder()
+            .id(UserCaseloadRoleIdentity.builder().caseload("NWEB").roleId(ROLE_ID).username(username).build())
+            .role(Role.builder().code("ROLE1").id(ROLE_ID).build())
+            .build()))
+        .accountDetail(buildAccountDetail(username, AccountStatus.OPEN))
+        .build()
+  }
+
+  private fun buildStaff(): Staff {
```
petergphillips, line 442: nicer for kotlin single liners to take away the `{` and `return` and use `=` instead

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/verify/ResetPasswordServiceImpl.java
```diff
@@ -21,29 +22,40 @@
...
+
+import static uk.gov.justice.digital.hmpps.oauth2server.security.AuthSource.nomis;
 
 @Service
 @Slf4j
 @Transactional
 public class ResetPasswordServiceImpl implements ResetPasswordService, PasswordService {
 
+    @SuppressWarnings("SqlResolve")
+    private static final String EXISTING_EMAIL_SQL = "select distinct internet_address from internet_addresses i " +
```
petergphillips, line 39: bit naff this is duplicated - better to call `VerifyEmailService.getExistingEmailAddresses` instead I think (unless we start getting into circular dependency hell)

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/verify/ResetPasswordServiceImpl.java
```diff
@@ -61,8 +69,8 @@ public ResetPasswordServiceImpl(final UserRepository userRepository,
 
     @Override
     @Transactional(transactionManager = "authTransactionManager")
-    public Optional<String> requestResetPassword(final String usernameOrEmailAddress, final String url) throws NotificationClientRuntimeException {
-        final Optional<User> optionalUser;
```
petergphillips, line 65: should be `final`

PR #478, opened 2020-06-09
DT-906: pathfinder national

petergphillips:


src/main/resources/db/auth/V4_01__national_pathfinder_group.sql
```diff
@@ -0,0 +1,14 @@
+-- only insert role if doesn't exist already
+INSERT INTO roles (role_id, role_code, role_name)
+SELECT NEWID(), 'PF_POLICE', 'Pathfinder Police'
```
petergphillips, line 3: most envs have the `PF_POLICE` role already, but this will ensure that it gets added to t2 too

PR #479, opened 2020-06-10
DT-896: Fix defaults for client details

petergphillips:


src/main/resources/db/auth/V4_03__default_client_details_jwt_fields.sql
```diff
@@ -0,0 +1,7 @@
+UPDATE oauth_client_details
+SET additional_information = '{"jwtFields":"-name,+user_name"}'
+WHERE additional_information = '{}';
```
petergphillips, line 3: so all existing data has values of `{}` since it has been persisted by the front end.  I assumed that it was `null`, but that was only dodgy test data that had never been saved

PR #484, opened 2020-06-12
DT-915: Configure timeouts and validate for token verification

petergphillips:


src/main/java/uk/gov/justice/digital/hmpps/oauth2server/config/TokenVerificationRestTemplateConfiguration.kt
```diff
@@ -21,12 +26,20 @@ class TokenVerificationRestTemplateConfiguration(private val apiDetails: TokenVe
...
-      restTemplateBuilder
-          .rootUri(tokenVerificationEndpointUrl)
-          .setConnectTimeout(apiTimeout)
-          .setReadTimeout(apiTimeout)
-          .configure(OAuth2RestTemplate(apiDetails))
+  fun tokenVerificationRestTemplate(restTemplateBuilder: RestTemplateBuilder): OAuth2RestTemplate {
+    // load balancer has timeout of 4 minutes so set to one second less than that
+    val poolManager = PoolingHttpClientConnectionManager(60 * 4 - 1, SECONDS)
+    poolManager.validateAfterInactivity = 500
+    poolManager.maxTotal = 10
```
petergphillips, line 33: this is configured by the default HttpClientBuilder when creating one from scratch, so need to apply the same properties when doing manually

PR #486, opened 2020-06-12
DT-895: Add in null appender and log to both for tomcat

petergphillips:


src/main/resources/application-oracle.yml
```diff
@@ -6,9 +6,6 @@ spring:
         show_sql: false
         jdbc:
           fetch_size: 100
-  profiles:
-    include: logstash
```
petergphillips, line 10: not used anymore

src/main/resources/logback-spring.xml
```diff
@@ -7,13 +7,8 @@
...
     <springProperty scope="context" name="app" source="spring.application.name"/>
 
-    <appender name="consoleAppender" class="ch.qos.logback.core.ConsoleAppender">
-        <layout class="ch.qos.logback.classic.PatternLayout">
-            <Pattern>${LOG_PATTERN}</Pattern>
-        </layout>
-    </appender>
-
     <springProfile name="dev || stdout">
+        <appender name="consoleAppender" class="ch.qos.logback.core.helpers.NOPAppender" />
```
petergphillips, line 11: when running in dev we were getting some messages twice.  By adding a no-op appender we prevent the duplicates

PR #498, opened 2020-07-01
PAT-942: SOCU roles and groups, plus locally seeded data for users and client credentials

timharrison-moj:
* Flyway additions for SOCU roles, groups, assignable roles - ok to be rolled out to all environments.
* Locally seeded data for client credentials
* Locally seeded data to setup Nomis user SOC_PRISON_LOCAL
* Locally seeded data to setup Auth users SOC_POLICE_LOCAL_NE + SOC_POLICE_LOCAL_NW & group assignment.
* Small update to README.md for local run / integration test instructions.


src/main/resources/db/auth/V4_06__manage_soc_cases_roles_groups.sql
```diff
@@ -0,0 +1,19 @@
+INSERT INTO roles (role_id, role_code, role_name) values (NEWID(), 'SOC_CUSTODY', 'SOCU Prison Role');
```
petergphillips, line 1: why is the code `SOC` but the description `SOCU`?

src/main/resources/db/dev/data/auth/V900_0__oauth_sample_data.sql
```diff
@@ -24,6 +24,8 @@ VALUES ('omicuser','1200','{"jwtFields":"-user_name"}','SYSTEM_READ_ONLY','passw
        ('whereabouts-api-client','3600','{}','ROLE_PAY, ROLE_CASE_NOTE_ADMIN,GLOBAL_SEARCH','client_credentials',null,'$2a$10$lBwbziQlLfiCnn8Kj1PfMujEcLdsJYlYSNJvBRO638gCYTS9yN0xm',null,null,'read,write',null),
        ('pathfinder-client','3600','{}',null,'authorization_code,refresh_token','read,write','$2a$10$WzgtydqXSuhdivpWDR3WXO.yjLBm4yuDqP64Og.7E4XURdrSfhOTi',43200,null,'read,write','http://localhost:3000/login/callback,http://localhost:3000'),
        ('pathfinder-admin','3600','{}','ROLE_GLOBAL_SEARCH,ROLE_SYSTEM_USER,ROLE_COMMUNITY','client_credentials','read,write','$2a$10$ajGimbJNWF1/FmZQMJWvieeQ/OdYaxWHQPgOjYDvvWu/4/744Yw7S',43200,null,'read,write',null),
+       ('manage-soc-cases-client','3600','{}',null,'authorization_code,refresh_token','read,write','$2a$10$WzgtydqXSuhdivpWDR3WXO.yjLBm4yuDqP64Og.7E4XURdrSfhOTi',43200,null,'read,write','http://localhost:3000/login/callback,http://localhost:3000'),
```
petergphillips, line 27: don't really want to grant `refresh_token`s these days, especially to new apps

README.md
```diff
@@ -3,7 +3,20 @@
...
+```bash
+SPRING_PROFILES_ACTIVE=dev ./gradlew bootRun
+```
+
+The service should start up using the dev profile, perform the flyway migrations on its local HSQLDB and then seed local development data.
+Can then be accessed in a browser on http://localhost:8080/auth/login
+
+### Run integration tests locally
+```bash
+SPRING_PROFILES_ACTIVE=dev,token-verification ./gradlew testFluentIntegration
```
petergphillips, line 18: don't need the active profile for running the tests.  However will need to start the app with
```bash
SPRING_PROFILES_ACTIVE=dev,token-verification ./gradlew bootRun
```
I think that is what you meant?

src/main/resources/db/dev/data/auth/V900_0__oauth_sample_data.sql
```diff
@@ -46,6 +48,7 @@ VALUES ('BOOK_MOVE', 'Book a secure move', 'Book a secure move', 'ROLE_PECS_SUPP
        ('OAUTHADMIN', 'Oauth Client Management', 'Manage Client Credentials for OAUTH2 Clients', 'ROLE_OAUTH_ADMIN', 'http://localhost:8080/auth/ui/', 1, null),
        ('POM', 'Allocate a POM Service', 'Allocate the appropriate offender manager to a prisoner', 'ROLE_ALLOC_MGR', 'https://moic.service.justice.gov.uk', 1, 'https://moic.service.justice.gov.uk/help'),
        ('PATHFINDER', 'Pathfinder Service', 'View and Manage Pathfinder nominals', 'ROLE_PF_STD_PRISON,ROLE_PF_APPROVAL,ROLE_PF_POLICE', 'http://localhost:3000', 1, null),
+       ('SOCU', 'Manage SOC Cases', 'View and Manage SOC cases', 'ROLE_SOC_CUSTODY,ROLE_SOC_COMMUNITY,ROLE_SOC_RESTRICTED', 'http://localhost:3000', 1, null),
```
petergphillips, line 51: just for info, Andy H doesn't like descriptions so has been taking them off.  Also names should be lower after initial title e.g. _Manage SOC cases_.  Wondering if it would be good to put the production details in the flyway script and then we can modify the URLs etc for each environment except production?  Also is there no feedback url yet?

src/main/resources/db/auth/V4_06__manage_soc_cases_roles_groups.sql
```diff
@@ -0,0 +1,19 @@
...
+INSERT INTO roles (role_id, role_code, role_name) values (NEWID(), 'SOC_COMMUNITY', 'SOCU Probation Role');
```
petergphillips, line 2: shouldn't this be `SOC_PROBATION` as we are trying to use probation instead of community?

src/main/resources/db/auth/V4_06__manage_soc_cases_roles_groups.sql
```diff
@@ -0,0 +1,19 @@
...
+INSERT INTO roles (role_id, role_code, role_name) values (NEWID(), 'SOC_RESTRICTED', 'SOCU Restricted Caseload');
```
petergphillips, line 3: so am I right in thinking all three of these roles can be assigned to auth users?  It looks like one is for NOMIS users, one for DELIUS and the other for auth, if that is the case then only need the `SOC_RESTRICTED` in the auth database otherwise people will start assigning it...

src/main/resources/db/dev/data/auth/V900_0__oauth_sample_data.sql
timharrison-moj: Ah yes, its been removed from Pathfinder in [dev, pre, prod] so will take it out locally here from both.

README.md
timharrison-moj: Think Circle does this to run the fluent integration tests too?

src/main/resources/db/auth/V4_06__manage_soc_cases_roles_groups.sql
timharrison-moj: I'll update the description here to match. The team is the Serious Organised Crime Unit and the service is manage SOC cases, so roles ought to relate to the service.

src/main/resources/db/dev/data/auth/V900_0__oauth_sample_data.sql
timharrison-moj: Happy to adjust description case but not sure about removing it.  Yes, good idea to add the production menu details - even though its only in dev at present, so nobody will actually have these roles or see it beyond dev. I'll add.

src/main/resources/db/auth/V4_06__manage_soc_cases_roles_groups.sql
timharrison-moj: The conversations have been going the other way in SOCU ! These were presented to them as SOC_PROBATION and SOC_PRISON, and they came back and said they would prefer SOC_CUSTODY and SOC_COMMUNITY - maybe that's the way they refer to offenders themselves? Not that they would even see these role codes.

src/main/resources/db/auth/V4_06__manage_soc_cases_roles_groups.sql
timharrison-moj: For these users, you're right. Its not quite that straightforward for what is planned, unfortunately.  There are another 2 or 3 types of user coming, and some will be AUTH-only (we think). These HQ and national users may need a mixture of CUSTODY and COMMUNITY roles - depending on their focus. The thinking is to make the small set of SOC roles available in auth to cover eventualities - I can remove them if the new users turn out to be just Delius and Nomis accounts. There's also at least one role that we'll map from a Delius role - ROLE_SOC_COMMUNITY - so does AUTH need to `know` about these?

src/main/resources/db/auth/V4_06__manage_soc_cases_roles_groups.sql
timharrison-moj: Within the team its variously referred to as SOCU and SOC - so probably a timely reminder for us to pick a term and stick to it!

README.md
petergphillips: so should be two commands i.e.
```bash
SPRING_PROFILES_ACTIVE=dev,token-verification ./gradlew bootRun
```
to start up auth in dev and token verification enabled

then to run the tests
```bash
./gradlew testFluentIntegration
```
so the spring profile not enabled when running the tests, only needed when running the application

src/main/resources/db/auth/V4_06__manage_soc_cases_roles_groups.sql
petergphillips: Fine. if you add the role to auth then it will appear on the dropdowns and can be assigned to users.  If it is a NOMIS or Delius role then it doesn't need to be in the auth `roles` table at all.

src/main/resources/db/dev/data/auth/V900_0__oauth_sample_data.sql
petergphillips: so re descriptions, worth pinging Andy Hudson, he's removed ours for _Manage account details_ and also _Manage user accounts_.
Didn't realise you only wanted dev at present.  I did come up with a cunning plan for inserting the oauth service records, write some dml like
```sql
insert to oauth_service (code, name...) select ('SOCU', 'Manage SOC Cases'...) from oauth_service where code = 'DETAILS' and url like '%dev%'
```
that will only write the row for dev and you could have two more insert statements for other environments when required.

README.md
timharrison-moj: OK thanks. Updated the README to reflect this.

src/main/resources/db/dev/data/auth/V900_0__oauth_sample_data.sql
timharrison-moj: I checked the service descriptions with Eddie and he said to keep it for SOC. The idea of removing descriptions was just for services that really don't need further explanation. I also tried a few variations of selectively adding the oauth_service dependent on DEV auth presence (similar to your example above) but the roles with commas, even escaped in various ways, were producing the column count error locally. I thought I saw a ticket recently for a form to manage oauth services in the environments, which might be a better bet?

timharrison-moj: Sorted the selective oauth_service entry now.

src/main/resources/db/dev/data/auth/V900_0__oauth_sample_data.sql
timharrison-moj: Done now.

src/main/resources/db/auth/V4_06__manage_soc_cases_roles_groups.sql
timharrison-moj: Understood. I think its right now - for what we know so far.

PR #499, opened 2020-07-06
DT-974: Add mfa to all manage user account roles

petergphillips:

PR #501, opened 2020-07-08
Revert "DT-974: Add mfa to all manage user account roles"

petergphillips:
Reverts ministryofjustice/hmpps-auth#499


helm_deploy/hmpps-auth/templates/_envs.tpl
```diff
@@ -166,11 +166,7 @@ env:
       secretKeyRef:
         key: TOKENVERIFICATION_ENDPOINT_URL
         name: {{ template "app.name" . }}
-
   - name: APPLICATION_AUTHENTICATION_MFA_WHITELIST
     value: {{ include "app.joinListWithComma" .Values.mfa_whitelist | quote }}
-
-  - name: APPLICATION_AUTHENTICATION_MFA_ROLES
-    value: "ROLE_MFA,ROLE_MAINTAIN_ACCESS_ROLES,ROLE_MAINTAIN_ACCESS_ROLES_ADMIN,ROLE_MAINTAIN_OAUTH_USERS,ROLE_AUTH_GROUP_MANAGER"
```
Mjwillis, line 174: ROLE_MFA should that still be there?

petergphillips: tbh it is the default in application.yml so the whole block can be removed

PR #502, opened 2020-07-08
DT-911: services dashboard 2

petergphillips:


src/main/java/uk/gov/justice/digital/hmpps/oauth2server/auth/model/Service.java
```diff
@@ -65,7 +64,15 @@ public String getDescription() {
     }
 
     public String getAuthorisedRoles() {
-        return this.authorisedRoles;
+        return authorisedRoles;
+    }
+
+    public String getAuthorisedRolesWithNewlines() {
+        return StringUtils.defaultIfBlank(authorisedRoles, "").replaceAll(",", "\n");
```
petergphillips, line 71: add in newline getter and setter so that we can have a text area rather than input for editing roles

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/resource/ServicesController.kt
```diff
@@ -1,15 +1,62 @@
...
 
 @Controller
 @RequestMapping("ui/services")
-class ServicesController(private val authServicesService: AuthServicesService) {
+class ServicesController(private val authServicesService: AuthServicesService,
+                         private val telemetryClient: TelemetryClient) {
 
   @GetMapping
-  fun userIndex() = ModelAndView("ui/services", mapOf("serviceDetails" to authServicesService.list()))
+  fun userIndex() = ModelAndView("ui/services", "serviceDetails", authServicesService.list())
```
petergphillips, line 24: use simple constructor

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/resource/ServicesController.kt
```diff
@@ -1,15 +1,62 @@
...
+    } else {
+      authServicesService.updateService(service)
+      telemetryClient.trackEvent("AuthServiceDetailsUpdate", telemetryMap, null)
+    }
+    return "redirect:/ui/services"
+  }
+
+  @GetMapping("/{code}/delete")
+  @PreAuthorize("hasRole('ROLE_OAUTH_ADMIN')")
+  fun deleteService(authentication: Authentication, @PathVariable code: String): String? {
```
petergphillips, line 55: like in /ui the delete is hidden behind a GET request so can be called from the browser

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/resource/UiController.kt
```diff
@@ -0,0 +1,14 @@
...
+import org.springframework.stereotype.Controller
+import org.springframework.web.bind.annotation.GetMapping
+import org.springframework.web.servlet.ModelAndView
+
+@Suppress("DEPRECATION")
+@Controller
+class UiController(private val clientDetailsService: JdbcClientDetailsService) {
+
+  @GetMapping("/ui")
+  fun userIndex() = ModelAndView("ui/index", "clientDetails", clientDetailsService.listClientDetails())
```
petergphillips, line 13: couldn't resist halving the size of the file by switching to kotlin

src/main/resources/templates/ui/services.html
```diff
@@ -19,22 +19,30 @@ <h2 class="govuk-heading-m">Services</h2>
...
-      <tr th:each="service : ${serviceDetails}" class="govuk-table__row">
+      <tr th:each="service : ${serviceDetails}" class="govuk-table__row" th:attr="data-qa=${service.code}">
         <td th:text="${service.name}" class="govuk-table__cell"></td>
         <td th:text="${service.description}" class="govuk-table__cell"></td>
         <td th:text="${service.roles}" class="govuk-table__cell"></td>
         <td class="govuk-table__cell">
           <span th:text="${service.url}"></span>
-          <span th:text="${' / ' + service.email}" th:if="${service.email}"></span>
-          <span th:text="${'/ no email set'}" th:unless="${service.email}"></span>
+          <span th:if="${#strings.isEmpty(service.email)}" th:text="${'/ no email set'}"></span>
```
petergphillips, line 32: improve code - empty string was being treated as being set

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/auth/model/Service.java
Mjwillis: Be good to do this with the client screen too at some point, on roles and callback urls making them text areas instead

PR #512, opened 2020-07-28
DT-953,DT-954: 🐛 Fix liveness and readiness probes

petergphillips:


src/main/resources/application.yml
```diff
@@ -50,15 +50,19 @@ management:
       base-path: /
       exposure:
         include: 'info, health, ping'
-  health:
-    probes:
-      enabled: true
```
petergphillips, line 55: this would still work, but has been deprecated and moved to under `endpoint`

PR #517, opened 2020-08-07
add skeleton code to modify user info based on "delius" scope

tomsmyers:
this PR takes a minor step towards delius integration with HMPPS auth via azure AD by conditionally modifying the subject, username, auth_source and user_id claims if the "delius" scope is granted to the client. currently the code does nothing, since the only supported mapping is from "azure" auth source, and even that isn't implemented fully.

if you were to put this code in production, and a client managed to get successfully request the `delius` scope and logged in using auth or nomis credentials, you would see an error like this in the logs for each token request:

```
uk.gov.justice.digital.hmpps.oauth2server.service.UserMappingException: auth source 'auth' not supported
        at uk.gov.justice.digital.hmpps.oauth2server.service.UserMappingService.map(UserMappingService.kt:22)
        at uk.gov.justice.digital.hmpps.oauth2server.config.JWTTokenEnhancer.enhance(JWTTokenEnhancer.java:70)
        ...
```

see https://dsdmoj.atlassian.net/wiki/spaces/AC/pages/2210955287/ADR-003+Differentiate+User+IDs+in+HMPPS+Auth+Tokens for more background on why we are modifying the user details.

### DIFF
--- a/src/main/java/uk/gov/justice/digital/hmpps/oauth2server/config/JWTTokenEnhancer.java
+++ b/src/main/java/uk/gov/justice/digital/hmpps/oauth2server/config/JWTTokenEnhancer.java
@@ -1,6 +1,7 @@
 package uk.gov.justice.digital.hmpps.oauth2server.config;
 
 
+import lombok.extern.slf4j.Slf4j;
 import org.apache.commons.lang3.StringUtils;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.security.core.GrantedAuthority;
@@ -12,6 +13,8 @@ import org.springframework.security.oauth2.provider.client.JdbcClientDetailsServ
 import org.springframework.security.oauth2.provider.token.TokenEnhancer;
 import uk.gov.justice.digital.hmpps.oauth2server.security.ExternalIdAuthenticationHelper;
 import uk.gov.justice.digital.hmpps.oauth2server.security.UserPersonDetails;
+import uk.gov.justice.digital.hmpps.oauth2server.service.UserContextService;
+import uk.gov.justice.digital.hmpps.oauth2server.service.UserMappingException;
 
 import java.util.Collections;
 import java.util.HashMap;
@@ -25,6 +28,7 @@ import java.util.stream.Stream;
 import static java.util.function.Predicate.not;
 
 @SuppressWarnings("deprecation")
+@Slf4j
 public class JWTTokenEnhancer implements TokenEnhancer {
     private static final String ADD_INFO_AUTH_SOURCE = "auth_source";
     static final String ADD_INFO_NAME = "name";
@@ -39,6 +43,9 @@ public class JWTTokenEnhancer implements TokenEnhancer {
     @Autowired
     private ExternalIdAuthenticationHelper externalIdAuthenticationHelper;
 
+    @Autowired
+    private UserContextService userContextService;
+
     @Override
     public OAuth2AccessToken enhance(final OAuth2AccessToken accessToken, final OAuth2Authentication authentication) {
         final Map<String, Object> additionalInfo;
@@ -46,23 +53,33 @@ public class JWTTokenEnhancer implements TokenEnhancer {
         if (authentication.isClientOnly()) {
             additionalInfo = addUserFromExternalId(authentication, accessToken);
         } else {
-            final var userAuthentication = authentication.getUserAuthentication();
-            final var userDetails = (UserPersonDetails) userAuthentication.getPrincipal();
-            final var userId = StringUtils.defaultString(userDetails.getUserId(), userAuthentication.getName());
-
             final var clientDetails = clientsDetailsService.loadClientByClientId(authentication.getOAuth2Request().getClientId());
+            final var userDetails = getUser(authentication);
+
             // note that DefaultUserAuthenticationConverter will automatically add user_name to the access token, so
             // removal of user_name will only affect the authorisation code response and not the access token field.
             additionalInfo = filterAdditionalInfo(
-                    Map.of(SUBJECT, userAuthentication.getName(),
+                    Map.of(SUBJECT, userDetails.getUsername(),
                             ADD_INFO_AUTH_SOURCE, StringUtils.defaultIfBlank(userDetails.getAuthSource(), "none"),
-                            ADD_INFO_USER_NAME, userAuthentication.getName(),
-                            ADD_INFO_USER_ID, userId,
-                            ADD_INFO_NAME, userDetails.getName()), clientDetails);
+                            ADD_INFO_USER_NAME, userDetails.getUsername(),
+                            ADD_INFO_USER_ID, userDetails.getUserId(),
+                            ADD_INFO_NAME, userDetails.getName()),
+                    clientDetails);
         }
 
         ((DefaultOAuth2AccessToken) accessToken).setAdditionalInformation(additionalInfo);
         return accessToken;
+    }
+
+    private UserPersonDetails getUser(final OAuth2Authentication authentication) {
+        final var loginUser = (UserPersonDetails) authentication.getUserAuthentication().getPrincipal();
+        try {
+            return userContextService.resolveUser(loginUser, authentication.getOAuth2Request().getScope());
+        } catch (UserMappingException e) {
+            log.error("user mapping failed", e);
+        }
+
+        return loginUser;
     }
 
     private Map<String, Object> filterAdditionalInfo(final Map<String, Object> info, final ClientDetails clientDetails) {
--- a/src/main/java/uk/gov/justice/digital/hmpps/oauth2server/resource/api/UserController.java
+++ b/src/main/java/uk/gov/justice/digital/hmpps/oauth2server/resource/api/UserController.java
@@ -72,16 +72,11 @@ public class UserController {
             @ApiResponse(code = 200, message = "OK", response = EmailAddress.class),
             @ApiResponse(code = 204, message = "No content.  No verified email address found for user"),
             @ApiResponse(code = 404, message = "User not found.  The user doesn't exist in auth so could have never logged in", response = ErrorDetail.class)})
-    public ResponseEntity<Object> getUserEmail(@ApiParam(value = "The username of the user.", required = true) @PathVariable final String username) {
-        final var user = userService.findUser(username);
-
-        if (user.isEmpty()) {
-            return notFoundResponse(username);
-        }
-
-        final var email = user.get();
-
-        return email.isVerified() ? ResponseEntity.ok(EmailAddress.fromUser(email)) : ResponseEntity.noContent().build();
+    public ResponseEntity<?> getUserEmail(@ApiParam(value = "The username of the user.", required = true) @PathVariable final String username) {
+        return userService
+                .getOrCreateUser(username)
+                .map(email -> email.isVerified() ? ResponseEntity.ok(EmailAddress.fromUser(email)) : ResponseEntity.noContent().build())
+                .orElseGet(() -> notFoundResponse(username));
     }
 
 
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 .gradle
 /build/
 !gradle/wrapper/gradle-wrapper.jar
+.DS_Store
 
 ### STS ###
 .apt_generated
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@
 /build/
 !gradle/wrapper/gradle-wrapper.jar
 .DS_Store
+.run/
 
 ### STS ###
 .apt_generated
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -21,6 +21,9 @@ dependencies {
 
   implementation("org.springframework.boot:spring-boot-starter-data-jpa")
 
+  implementation("org.springframework.boot:spring-boot-starter-oauth2-client")
+  implementation("org.springframework.boot:spring-boot-starter-webflux")
+
   implementation("javax.annotation:javax.annotation-api:1.3.2")
   implementation("javax.xml.bind:jaxb-api:2.3.1")
   implementation("com.sun.xml.bind:jaxb-impl:2.3.3")
--- a/src/main/java/uk/gov/justice/digital/hmpps/oauth2server/azure/AzureUserPersonDetails.kt
+++ b/src/main/java/uk/gov/justice/digital/hmpps/oauth2server/azure/AzureUserPersonDetails.kt
@@ -1,0 +1,56 @@
+package uk.gov.justice.digital.hmpps.oauth2server.azure
+
+import org.springframework.security.core.GrantedAuthority
+import uk.gov.justice.digital.hmpps.oauth2server.auth.model.Person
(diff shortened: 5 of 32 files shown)


src/main/java/uk/gov/justice/digital/hmpps/oauth2server/config/JWTTokenEnhancer.java
```diff
@@ -53,12 +54,24 @@ public OAuth2AccessToken enhance(final OAuth2AccessToken accessToken, final OAut
...
+                ADD_INFO_USER_NAME, userAuthentication.getName(),
+                ADD_INFO_USER_ID, userId,
+                ADD_INFO_NAME, userDetails.getName()
+            ));
+
+            // this is a temporary change to enable initial delius integration.
+            // in the future tokens will contain either "delius_id", "nomis_id",
+            // "oasys_id" etc. depending on the client that requested the token
+            if (userDetails.getAuthSource().equals("delius") &&
+                clientDetails.getAuthorities().stream().anyMatch(a -> a.getAuthority().equals("ROLE_ACCESS_DELIUS_ID"))) {
```
Mjwillis, line 70: Isn't there a problem where if this is a "authorisation_code" grant type the authorities are not included in the JWT but are taken from the users account not the values in the client cred setup

tomsmyers: thanks for the comment, i've been testing with auth code grant types and not seen this issue, but i'll dig into it.

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/config/JWTTokenEnhancer.java
```diff
@@ -41,7 +42,7 @@
 
     @Override
     public OAuth2AccessToken enhance(final OAuth2AccessToken accessToken, final OAuth2Authentication authentication) {
-        final Map<String, Object> additionalInfo;
+        Map<String, Object> additionalInfo;
```
petergphillips, line 45: styling point - all variables should be immutable, would be good for contents of maps to be immutable too

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/config/JWTTokenEnhancer.java
petergphillips: the `clientDetails` is the results of loading the client by client id from the database - note not using the user details authorities, so will be fine

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/resource/api/UserController.java
```diff
@@ -33,10 +36,26 @@
...
             @ApiResponse(code = 401, message = "Unauthorized", response = ErrorDetail.class)})
-    public UserDetail me(@ApiIgnore final Principal principal) {
+    public UserDetail me(@ApiIgnore final Principal principal,
+                         @ApiIgnore final Authentication authentication) {
         final var user = userService.findMasterUserPersonDetails(principal.getName());
+        final var userDetail = user.map(UserDetail::fromPerson).orElse(UserDetail.fromUsername(principal.getName()));
 
-        return user.map(UserDetail::fromPerson).orElse(UserDetail.fromUsername(principal.getName()));
+        // if additional user IDs are present in the access token, include them in the UserDetail object
+        String deliusId = null;
```
petergphillips, line 45: prefer immutable variables

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/resource/api/UserController.java
```diff
@@ -33,10 +36,26 @@
...
+        final String token = ((OAuth2AuthenticationDetails)authentication.getDetails()).getTokenValue();
+        try {
+            final var tokenClaims = JWTParser.parse(token).getJWTClaimsSet();
```
petergphillips, line 48: Odd that you need to do this.  I would have thought that the JWT has already been parsed and would be in the authentication.  I'll take a look...

tomsmyers: i thought the same... but i couldn't find the decoded token or claims anywhere in the Authentication object

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/config/JWTTokenEnhancer.java
petergphillips: It's sort of a hack of the roles, you're using a role to indicate that a field should be added to the JWT.  We allow customisation of the JWT via the `jwtFields` configuration already, looks like it would be better placed in there so add in `+delius_usernames` to that for the client.
That said, access tokens are often passed around from system to system.  The `jwtFields` approach only works if it is a front end client that requires it since if prison-api required it then it would have to be added to all systems that could call prison-api.

tomsmyers: personally, i wouldn't call it a hack. it's just one way to configure a client to get a certain additional claim in the token. we identified 3 ways you could configure this using the existing hmpps auth managment UI. 1) roles 2) scopes 3) jwt fields. i'm not fanatical about any in particular but there should be justification for whichever we choose. 

i think the role based config is easy to understand; "clients with roles `ACCESS_DELIUS_ID` get delius ids in the token". same with scopes; "clients with access to scope `delius_id` get delius ids in the token". in fact scopes are more commonly used to "request claims" so this may even be a better fit. the jwt fields based config also does what it says on the tin: "clients with `+delius_id` in the jwt fields configuration get delius ids in the token". however i think it's less configurable e.g. a scenario like "clients with access to scope `delius_profile` get the following claims in the token..."

petergphillips: Apologies, hack was the wrong word, which I think I realised when I called it a _sort of a hack_.  In my mind roles are normally used to check what a user is allowed to see.

tomsmyers:
@andymarke this stuff is somewhat related to a recent thread on slack which is totally unconnected to this change (https://mojdt.slack.com/archives/CK0QLCQ1G/p1596623024139600). 

i invited you to a discussion later today regarding these changes. mentioning you here so you have the context. also worth taking a skim of the ADR mentioned in main description.

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/resource/api/UserController.java
petergphillips: I've had another look.  Seems weird that after getting all the user details we then have to decode the token to retrieve other details.  That aside, rather than decoding the token again would be better to implement our own `UserAuthenticationConverter.extractAuthentication` which implemented our own `Authentication` (maybe extending `UsernamePasswordAuthenticationToken`) that had the necessary methods on it.
I'm still trying to get my head round a `/user/me` endpoint that returns different data depending on the client that asked for the token to be created.  Are my personas effectively then simplified to a single one by creating the token and this method then just returns the one that is in scope?

tomsmyers: happy with the refactoring idea.

this is a standard pattern in openid connect (albeit using scopes, not roles), where the userinfo response contains conditional claims depending on scopes granted to the client. an example in the google identity platform docs: https://developers.google.com/identity/protocols/oauth2/openid-connect#obtaininguserprofileinformation

in this specific scenario the reason we pluck the delius id from the token here is that we want the option to prompt the user during login in the event there is not a 1:1 mapping between email address and delius user id. by the time the client is calling this endpoint, it's too late.

petergphillips: Very interesting, scopes looks a nice clean way to do it :+1:

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/config/JWTTokenEnhancer.java
```diff
@@ -48,17 +55,33 @@ public OAuth2AccessToken enhance(final OAuth2AccessToken accessToken, final OAut
         } else {
             final var userAuthentication = authentication.getUserAuthentication();
             final var userDetails = (UserPersonDetails) userAuthentication.getPrincipal();
-            final var userId = StringUtils.defaultString(userDetails.getUserId(), userAuthentication.getName());
 
             final var clientDetails = clientsDetailsService.loadClientByClientId(authentication.getOAuth2Request().getClientId());
             // note that DefaultUserAuthenticationConverter will automatically add user_name to the access token, so
             // removal of user_name will only affect the authorisation code response and not the access token field.
+
+            var authSource = StringUtils.defaultIfBlank(userDetails.getAuthSource(), "none");
```
tomsmyers, line 63: @petergphillips  how would you structure this code to avoid mutable vars here?

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/service/UserMappingService.kt
```diff
@@ -0,0 +1,29 @@
...
+import uk.gov.justice.digital.hmpps.oauth2server.auth.model.User
+import uk.gov.justice.digital.hmpps.oauth2server.delius.service.DeliusUserService
+import uk.gov.justice.digital.hmpps.oauth2server.maintain.AuthUserService
+import uk.gov.justice.digital.hmpps.oauth2server.security.UserPersonDetails
+
+class UserMappingException(message: String): Exception(message)
+
+@Service
+class UserMappingService {
+    companion object {
```
petergphillips, line 15: should be 2 character indent for kotlin

tomsmyers: is there a lint or format task i can run to pick up style rules like this?

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/config/JWTTokenEnhancer.java
petergphillips: well first stab would be something like:
```java
            final var loginAuthSource = StringUtils.defaultIfBlank(userDetails.getAuthSource(), "none");
            final var loginUserName = userAuthentication.getName();
            final var loginUserId = StringUtils.defaultString(userDetails.getUserId(), userAuthentication.getName());

            final String authSource;
            final String userName;
            final String userId;
            if (authentication.getOAuth2Request().getScope().contains("delius")) {
                // client has been granted 'delius' scope; adjust the user info if possible
                final var deliusUser = userMappingService.map(loginUserName, loginAuthSource, "delius");
                authSource = deliusUser.getAuthSource();
                userName = deliusUser.getUsername();
                userId = deliusUser.getUserId();
            } else {
                authSource = loginAuthSource;
                userName = loginUserName;
                userId = loginUserId;
            }
```
which is a bit naff, but makes it very clear what is going on with the mapping.  Intellij refactor out method then gives
```java
            final var loginAuthSource = StringUtils.defaultIfBlank(userDetails.getAuthSource(), "none");
            final var loginUserName = userAuthentication.getName();
            final var loginUserId = StringUtils.defaultString(userDetails.getUserId(), userAuthentication.getName());
            final var authUserMapper = new AuthUserMapper(authentication, loginAuthSource, loginUserName, loginUserId);

            additionalInfo = filterAdditionalInfo(
                    Map.of(SUBJECT, authUserMapper.getUserName(),
                            ADD_INFO_AUTH_SOURCE, authUserMapper.getAuthSource(),
                            ADD_INFO_USER_NAME, authUserMapper.getUserName(),
                            ADD_INFO_USER_ID, authUserMapper.getUserId(),
                            ADD_INFO_NAME, userDetails.getName()), clientDetails);
```
which is a lot nicer and much shorter, but still not happy.

I can't help but think that we are hard coding `delius` into the token enhancer, so maybe the mapper would be better with that logic.  We would then have:
```java
            final var user = userMappingService.map(clientDetails.getScope(), userDetails);

            additionalInfo = filterAdditionalInfo(
                    Map.of(SUBJECT, user.getUsername(),
                            ADD_INFO_AUTH_SOURCE, user.getAuthSource(),
                            ADD_INFO_USER_NAME, user.getUsername(),
                            ADD_INFO_USER_ID, user.getUserId(),
                            ADD_INFO_NAME, userDetails.getName()), clientDetails);
```
i.e. we always call the mapper with the scopes and it is up to that service to then map from the provided scopes to the relevant type of user, or it is a no-op and returns the `userDetails` that is passed in...

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/config/JWTTokenEnhancer.java
```diff
@@ -48,17 +55,35 @@ public OAuth2AccessToken enhance(final OAuth2AccessToken accessToken, final OAut
...
+                }
+            }
+
             additionalInfo = filterAdditionalInfo(
-                    Map.of(SUBJECT, userAuthentication.getName(),
-                            ADD_INFO_AUTH_SOURCE, StringUtils.defaultIfBlank(userDetails.getAuthSource(), "none"),
-                            ADD_INFO_USER_NAME, userAuthentication.getName(),
-                            ADD_INFO_USER_ID, userId,
-                            ADD_INFO_NAME, userDetails.getName()), clientDetails);
+                Map.of(SUBJECT, userName,
```
petergphillips, line 82: interesting re intellij - I was beginning to suspect you weren't using it since the change to the indenting here is odd :grin:

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/config/JWTTokenEnhancer.java
tomsmyers: 👍 i'll have a play around with these ideas.

on the code in 'first stab', you end up repeating the content of the 'else' block for the 'catch' block too, so then you have loads of repetition. 

on the last point, i'm happy with that idea, but then your mapping service is no longer really a mapping service, because it longer has the concept of "from" and "to". it's more like a "user context resolution service" or something 🤔

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/config/JWTTokenEnhancer.java
tomsmyers: i've opened this up in idea now and it's not complaining about the indentation here!

tomsmyers:
i think from my perspective the changes are 'done' at this point. i'll add some tests and remove draft status later today.

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/config/JWTTokenEnhancer.java
petergphillips: Not sure what you mean.  Intellij won't warn about formatting.  You can get intellij to format on save, but it doesn't do it by default.  You therefore have to reformat the file manually or on commit by ticking the box.

tomsmyers:
closing this PR for now until azure work is completed

PR #518, opened 2020-08-09
DCS-555 Automatically migrate users into auth on "/api/user/{username}/email" call 

bell-pa:


src/main/java/uk/gov/justice/digital/hmpps/oauth2server/security/UserService.kt
```diff
@@ -11,59 +11,79 @@ import uk.gov.justice.digital.hmpps.oauth2server.auth.model.User.EmailType
...
                   private val authUserService: AuthUserService,
                   private val deliusUserService: DeliusUserService,
-                  private val userRepository: UserRepository) {
+                  private val userRepository: UserRepository,
+                  private val verifyEmailService: VerifyEmailService) {
 
-  companion object {
-    val log: Logger = LoggerFactory.getLogger(this::class.java)
-  }
+    companion object {
```
petergphillips, line 26: should be 2 character indent for kotlin

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/security/UserService.kt
```diff
@@ -11,59 +11,79 @@ import uk.gov.justice.digital.hmpps.oauth2server.auth.model.User.EmailType
...
+            }
 
-  fun hasVerifiedMfaMethod(userDetails: UserPersonDetails): Boolean {
-    val user = findUser(userDetails.username).orElseGet { userDetails.toUser() }
-    return user.hasVerifiedMfaMethod()
-  }
+    fun getEmailAddressFromNomis(username: String): Optional<String> {
+        val emailAddresses = verifyEmailService.getExistingEmailAddresses(username)
+        val justiceEmail = emailAddresses
+                .stream()
```
petergphillips, line 65: don't need the `stream` or `collect` as using kotlin which supports it natively

src/test/java/uk/gov/justice/digital/hmpps/oauth2server/security/UserServiceTest.kt
```diff
@@ -20,224 +20,290 @@ import uk.gov.justice.digital.hmpps.oauth2server.maintain.AuthUserService
...
+    private val verifyEmailService: VerifyEmailService = mock()
+    private val userService = UserService(nomisUserService, authUserService, deliusUserService, userRepository, verifyEmailService)
 
-    @Test
-    fun `findMasterUserPersonDetails nomis user`() {
-      whenever(nomisUserService.getNomisUserByUsername(anyString())).thenReturn(staffUserAccountForBob)
-      val user = userService.findMasterUserPersonDetails("bob")
-      assertThat(user).isPresent.get().extracting { it.username }.isEqualTo("nomisuser")
-    }
+    //     val emailAddresses = mutableListOf("Bob.smith@hmps.gsi.gov.uk", "Bob.smith@justice.gov.uk", "Bob.smith2@justice.gov.uk")
```
petergphillips, line 35: please delete commented out code

src/test/java/uk/gov/justice/digital/hmpps/oauth2server/security/UserServiceTest.kt
```diff
@@ -20,224 +20,290 @@ import uk.gov.justice.digital.hmpps.oauth2server.maintain.AuthUserService
...
+    inner class GetOrCreateUserWithEmail {
 
-  @Nested
-  inner class IsSameAsCurrentVerifiedMobile {
+        @Test
+        fun `getOrCreateUserWithEmail user exists already`() {
+            val user = User.of("joe")
+            whenever(userRepository.findByUsername(anyString())).thenReturn(Optional.of(user))
+            val newUserOpt = userService.getOrCreateUser("joe")
+            // can't compare directly User.equals only compares id!
```
petergphillips, line 106: I think that is why the previous test did a `isSameAs` - does that not work?

src/test/java/uk/gov/justice/digital/hmpps/oauth2server/verify/ResetPasswordServiceTest.kt
```diff
@@ -9,8 +9,7 @@ import com.nhaarman.mockitokotlin2.whenever
 import org.assertj.core.api.Assertions.assertThat
 import org.assertj.core.api.Assertions.assertThatThrownBy
 import org.junit.jupiter.api.Test
-import org.mockito.ArgumentMatchers.anyMap
-import org.mockito.ArgumentMatchers.anyString
+import org.mockito.ArgumentMatchers.*
```
petergphillips, line 12: trying not to use wildcard imports these days

src/test/java/uk/gov/justice/digital/hmpps/oauth2server/verify/ResetPasswordServiceTest.kt
```diff
@@ -9,8 +9,7 @@ import com.nhaarman.mockitokotlin2.whenever
...
 import org.mockito.Mockito.verify
```
petergphillips, line 13: this is the wrong `verify` - should be from mokitokotlin2

src/test/java/uk/gov/justice/digital/hmpps/oauth2server/verify/ResetPasswordServiceTest.kt
```diff
@@ -36,434 +29,427 @@ import java.util.*
...
+
+    @Test
+    fun `Nomis User who has not logged into auth can reset password`() {
+        whenever(userRepository.findByUsername(anyString())).thenReturn(Optional.empty())
+        whenever(userService.findMasterUserPersonDetails(anyString())).thenReturn(Optional.of(buildStandardUser("user")))
+        whenever(userService.getEmailAddressFromNomis(anyString())).thenReturn(Optional.of("Bob.smith@justice.gov.uk"))
+        val optionalLink = resetPasswordService.requestResetPassword("user", "url")
+        assertThat(optionalLink).isPresent
+        verify(userRepository).save(argThat { user ->
+                    user.username == "user" &&
```
petergphillips, line 386: this look a bit naff.  If it doesn't match it will just return `false` and then we won't get an error message etc.
Better to write something like
```
    verify(userRepository).save(check<User> { user ->
      assertThat(user.username).isEqualTo("user2")
      ...
```

src/test/java/uk/gov/justice/digital/hmpps/oauth2server/security/UserServiceTest.kt
bell-pa: It is comparing different Optional instances so `isSameAs` doesn't work. 
This assertion `assertThat(newUserOpt).hasValue(User.of("fred"))` passes (though it shouldn't) because of the `equals` implementation on `User` so I think the assertion has to express its own version of what 'equals' means.

petergphillips: I more meant changing too
```
            assertThat(newUserOpt).get().isSameAs(user)
```
which I think should work and then will be simpler

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/security/UserService.kt
```diff
@@ -42,12 +44,28 @@ class UserService(private val nomisUserService: NomisUserService,
...
+          }
+          userRepository.save(user)
+        }
       }
 
+  fun getEmailAddressFromNomis(username: String): Optional<String> {
+    val emailAddresses = verifyEmailService.getExistingEmailAddresses(username)
+    val justiceEmail = emailAddresses
+        .filter { email -> email.endsWith("justice.gov.uk") }
+        .toList()
```
petergphillips, line 65: redundant `toList`?

src/test/java/uk/gov/justice/digital/hmpps/oauth2server/security/UserServiceTest.kt
```diff
@@ -55,24 +57,82 @@ class UserServiceTest {
...
+      whenever(verifyEmailService.getExistingEmailAddresses(any())).thenReturn(listOf("a@b.gov.uk"))
+      val optionalAddress = userService.getEmailAddressFromNomis("joe")
+      assertThat(optionalAddress).isEmpty
+    }
+
+    @Test
+    fun `getEmailAddressFromNomis one justice email`() {
+      whenever(verifyEmailService.getExistingEmailAddresses(any())).thenReturn(listOf("Bob.smith@hmps.gsi.gov.uk", "Bob.smith@justice.gov.uk"))
+      val optionalAddress = userService.getEmailAddressFromNomis("joe")
+      assertThat(optionalAddress).hasValue("Bob.smith@justice.gov.uk")
```
petergphillips, line 79: hadn't seen `hasValue` before in assertj :+1:

PR #520, opened 2020-08-13
Azure AD as an auth source, auth code flow

electrified:
This is a pull request that integrates Azure AD as an auth source into HMPPS Auth

I can provide config so this can be tried locally with the DEVL Azure AD instance

- The username stored in the username field is the Azure OID UUID. This requires increasing the username field length to 37 characters. I looked at using the id column, but that change would have been quite invasive, and we still need a unique value to put in the username column.

- Health check to follow in future PR as this is already large
- Changes to account-details page (potentially hide change email link etc) to also follow in future PR


src/main/java/uk/gov/justice/digital/hmpps/oauth2server/azure/service/AzureUserService.kt
```diff
@@ -0,0 +1,37 @@
...
+
+@Slf4j
+@Service
+class AzureUserService(private val userRepository: UserRepository) {
+
+    companion object {
+        val log: Logger = LoggerFactory.getLogger(this::class.java)
+    }
+
+    fun getAzureUserByUsername(username: String?): Optional<AzureUserPersonDetails> {
```
electrified, line 20: Currently this simply looks at the internal auth database for user lookups. 

Ideally we would determine the current account status by querying Azure either by:
- Retaining the users original Azure token, and using this (or refresh token) to call the Azure userinfo endpoint. Would require some fairly major reworking
- Using client credentials  get user info from Microsoft Graph API by email. This requires additional admin granted privileges on the client to read directory info. This has been investigated, but didn't figure out how to get account locked status

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/security/OidcJwtAuthenticationSuccessHandler.kt
```diff
@@ -0,0 +1,60 @@
...
+                                          @Qualifier("tokenVerificationApiRestTemplate") restTemplate: RestTemplate?,
+                                          @Value("\${tokenverification.enabled:false}") tokenVerificationEnabled: Boolean,
+                                          private val userRetriesService: UserRetriesService) : JwtAuthenticationSuccessHandler(jwtCookieHelper, jwtAuthenticationHelper, cookieRequestCache, verifyEmailService, restTemplate, tokenVerificationEnabled) {
+    @Throws(IOException::class, ServletException::class)
+    override fun onAuthenticationSuccess(request: HttpServletRequest, response: HttpServletResponse,
+                                         authentication: Authentication) {
+        val authenticationResult = authentication as OAuth2AuthenticationToken
+        val principal = authenticationResult.principal as DefaultOidcUser
+
+        // Enables compatibility with dev and production Azure configurations
```
electrified, line 32: Prod DOM1 is federated into the dev instance, so you can log in with prod or dev accounts. The token fields differ due to configuration however. This is to work around this

src/main/resources/application-dev-config.yml
```diff
@@ -2,13 +2,31 @@ spring:
   h2:
     console:
       enabled: true
+  security:
```
electrified, line 5: This could be defined in code rather than config?

src/main/resources/db/auth/V5_1__increase_username_length.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE user_retries ALTER COLUMN username TYPE varchar(240);
```
electrified, line 1: Column widths increased for email usernames

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/azure/AzureUserPersonDetails.kt
```diff
@@ -0,0 +1,56 @@
...
+    override fun getUserId(): String = username
+
+    override fun getName(): String = "$firstName $surname"
+
+    override fun getFirstName(): String = firstName
+
+    override fun getAuthSource(): String? = AuthSource.azure.name
+
+    override fun eraseCredentials() {
+    }
```
petergphillips, line 42: should really set `password` to `null` here, although obviously doesn't matter for this demo...

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/azure/AzureUserPersonDetails.kt
```diff
@@ -0,0 +1,56 @@
...
+import uk.gov.justice.digital.hmpps.oauth2server.security.UserPersonDetails
+
+data class AzureUserPersonDetails(private val authorities: MutableCollection<GrantedAuthority>,
+                                  private val enabled: Boolean,
+                                  private val username: String,
+                                  private val firstName: String,
+                                  val surname: String,
+                                  val email: String,
+                                  private val credentialsNonExpired: Boolean,
+                                  private val password: String,
```
tomsmyers, line 16: does this ever get constructed with a real password?

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/azure/service/AzureUserService.kt
tomsmyers, line 20: can this return an `AzureUserPersonDetails?` instead of java optional?

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/config/AuthenticationManagerConfiguration.java
```diff
@@ -128,6 +154,33 @@ protected void configure(final HttpSecurity http) throws Exception {
         // @formatter:on
     }
 
+    /**
+     * Custom User Service for the Azure OIDC integration. Spring expects to get the username from the userinfo endpoint,
```
tomsmyers, line 158: this is frustrating isn't it. guessing there's no way to configure the userinfo response to contain the fields we want in azure?

electrified: As far as I know, the contents of the userinfo response is based on the scopes in the token you call the endpoint with, with the proviso that the AD user has to have the data for those fields.

The response is controlled by these scopes:
  "openid",
  "profile",
  "email",
  "offline_access"

I've tried getting a token with those, and still didn't get any fields back with the email address in.

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/security/OidcJwtAuthenticationSuccessHandler.kt
```diff
@@ -0,0 +1,60 @@
...
+                                          jwtAuthenticationHelper: JwtAuthenticationHelper?,
+                                          cookieRequestCache: CookieRequestCache?,
+                                          verifyEmailService: VerifyEmailService?,
+                                          @Qualifier("tokenVerificationApiRestTemplate") restTemplate: RestTemplate?,
+                                          @Value("\${tokenverification.enabled:false}") tokenVerificationEnabled: Boolean,
+                                          private val userRetriesService: UserRetriesService) : JwtAuthenticationSuccessHandler(jwtCookieHelper, jwtAuthenticationHelper, cookieRequestCache, verifyEmailService, restTemplate, tokenVerificationEnabled) {
+    @Throws(IOException::class, ServletException::class)
+    override fun onAuthenticationSuccess(request: HttpServletRequest, response: HttpServletResponse,
+                                         authentication: Authentication) {
+        val authenticationResult = authentication as OAuth2AuthenticationToken
```
tomsmyers, line 29: this is a generic handler for OIDC auth success; is there ever a chance where these unsafe casts could not be correct?

src/main/resources/application-dev-config.yml
```diff
@@ -2,13 +2,31 @@ spring:
...
+            user-name-attribute: sub
+            jwk-set-uri: https://login.microsoftonline.com/0bb413d7-160d-4839-868a-f3d46537f6af/discovery/v2.0/keys
 
 application:
   smoketest:
     enabled: true
 
   authentication:
     lockout-count: 3
+    show-azure-link: true
```
tomsmyers, line 29: is there anyway to login to hmpps via azure when this is false e.g. manually crafting URLs ?

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/azure/service/AzureUserService.kt
electrified: I was following the pattern used in the DeliusUserService which also returns an optional.  Seemed best for consistency?

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/security/AuthSource.java
```diff
@@ -5,7 +5,7 @@
 import java.util.Optional;
 
 public enum AuthSource {
-    nomis, auth, delius, none;
+    nomis, auth, delius, azure, none;
```
petergphillips, line 8: calling it `azure` is confusing since the `auth` database is in `azure` too.  Isn't this actually `dom1`?

src/main/resources/db/auth/V5_1__increase_username_length.sql
petergphillips: Don't understand.  Is this suggesting that the username is actually the email address?

src/main/resources/templates/login.html
```diff
@@ -93,9 +93,12 @@ <h1 class="govuk-heading-l">Sign in</h1>
...
       </div>
     </div>
-    <div class="govuk-form-group">
-      <input class="govuk-button" id="submit" type="submit" role="button" data-element-id="continue-button"
-             value="Sign in">
+    <div th:if="${@environment.getProperty('application.authentication.show-azure-link')}"  class="govuk-form-group">
+      <ul class="govuk-list">
+      <li><input class="govuk-button" id="submit" type="submit" role="button" data-element-id="continue-button"
+             value="Sign in"></li>
+        <li><a class="govuk-link" href="/auth/oauth2/authorization/microsoft">Log in with a Windows DOM1 account</a></li>
```
petergphillips, line 100: would be interesting to find out what the users know this as - is it your windows credentials?

src/main/resources/db/auth/V5_1__increase_username_length.sql
electrified: Yes that's correct. Effectively this duplicates what's in the email field so it is a bit redundant.

Based on our conversation earlier around UUIDs/usernames etc I can revert this part of the change.

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/security/AuthSource.java
petergphillips: although after our discussions not sure what to call it - is it `azuread`?

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/security/JwtAuthenticationHelper.java
```diff
@@ -47,9 +49,25 @@ public JwtAuthenticationHelper(@Value("${jwt.signing.key.pair}") final String pr
...
+    String createJwtWithIdFromOidcAuthentication(final OAuth2AuthenticationToken authentication, final String jwtId) {
+        final var userDetails = (DefaultOidcUser) authentication.getPrincipal();
+        final var username = userDetails.getName().toUpperCase();
+        log.debug("Creating jwt cookie for user {}", username);
+        final var authorities = Optional.ofNullable(authentication.getAuthorities()).orElse(Collections.emptyList());
+        final var authoritiesAsString = authorities.stream().map(GrantedAuthority::getAuthority).collect(Collectors.joining(","));
+        return Jwts.builder()
+                .setId(jwtId)
+                .setSubject(username)
+                .addClaims(Map.of("authorities", authoritiesAsString, "name", userDetails.getFullName(), "auth_source", "azure", "user_id", userDetails.getPreferredUsername()))
```
tomsmyers, line 66: did we change the auth source to "azuread"?

src/main/resources/application-dev-config-azure-ad-test.yml
```diff
@@ -0,0 +1,25 @@
...
+            authorization-uri: http://localhost:8101/oauth2/v2.0/authorize
+            token-uri: http://localhost:8101/tenant-id/oauth2/v2.0/token
+            user-info-uri: http://localhost:8101/oidc/userinfo
+            user-name-attribute: sub
+            jwk-set-uri: http://localhost:8101/tenant-id/discovery/v2.0/keys
+
+application:
+  authentication:
+    microsoft:
+      linktext: Log in with a justice.gov.uk email
```
tomsmyers, line 25: 👍🏼

src/main/resources/templates/login.html
```diff
@@ -93,9 +93,12 @@ <h1 class="govuk-heading-l">Sign in</h1>
         </div>
       </div>
     </div>
-    <div class="govuk-form-group">
-      <input class="govuk-button" id="submit" type="submit" role="button" data-element-id="continue-button"
-             value="Sign in">
+    <div th:each="oauth2Client: ${oauth2Clients}"  class="govuk-form-group">
```
tomsmyers, line 96: if there were more than one client in this list, would this code create a green 'sign in' button alongside each link? is that the correct behaviour?

src/main/resources/templates/login.html
electrified: The link text is now configurable. Currently the text is configured to "Log in with your justice.gov.uk credentials"

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/security/JwtAuthenticationHelper.java
electrified: We did, thanks for spotting this.

src/main/resources/templates/login.html
electrified: Yes, this is wrong. Will fix

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/security/AuthSource.java
electrified: This has been renamed to `azuread`

build.gradle.kts
```diff
@@ -18,10 +18,12 @@ dependencies {
   implementation("org.springframework.security:spring-security-jwt:1.1.1.RELEASE")
   implementation("org.springframework.security.oauth:spring-security-oauth2:2.5.0.RELEASE")
   implementation("io.jsonwebtoken:jjwt:0.9.1")
-  implementation("com.nimbusds:nimbus-jose-jwt:9.0.1")
```
electrified, line 21: This overrides the older version the oauth2-client depends on, resulting in errors about methods not existing

PR #524, opened 2020-08-25
DT 1080: upgrade spring boot plugin to 1.0.0 and Kotlin to 1.4

SimonMitchellMOJ:


src/main/resources/application.yml
```diff
@@ -58,14 +58,14 @@ management:
...
         enabled: true
-      group:
-        liveness:
-          include: livenessStateProbeIndicator
-        readiness:
-          include: readinessStateProbeIndicator
     info:
       cache:
         time-to-live: 2000ms
+  health:
```
petergphillips, line 64: don't think you need this stuff any more

petergphillips: according to https://github.com/spring-projects/spring-boot/issues/23035 just need to remove the workarounds above

src/main/resources/application.yml
```diff
@@ -58,15 +58,11 @@ management:
       show-details: always
       probes:
         enabled: true
-      group:
```
andymarke, line 61: so I assume this is no longer needed?

andymarke (approved):
LGTM

src/main/resources/application.yml
SimonMitchellMOJ: It appears we needed it because of something they did with spring boot 2.2.2 but have reverted/fixed within spring boot 2.2.3

PR #525, opened 2020-08-25
take delius username from /users/{username}/details response instead of from login form input

tomsmyers:
As of https://github.com/ministryofjustice/community-api/pull/249 (prod version `2020-08-19.1895.8896360`), the community-api `/users/{username}/details` response contains the username as an additional field. 

This PR changes `DeliusUserService` to pull the username from the community-api response instead of using the value passed by the user. 

As part of this change, I took the opportunity to standardise the usernames used for Delius integration tests. All error conditions under test use `DELIUS_ERROR_…` as the username. The successful responses use wiremock’s `ResponseTemplateTransformer` to return the correct username in the response. 

The driver behind this change is to simplify the future changes required to map Azure AD users to Delius users by looking up the Delius user via email address. These changes were originally part of a different PR, but it made sense to pull them out and review separately.


.gitignore
```diff
@@ -1,6 +1,8 @@
 .gradle
 /build/
 !gradle/wrapper/gradle-wrapper.jar
+.DS_Store
```
petergphillips, line 4: don't amend, but in future these sort of things would be better going in your global .gitignore - see https://gist.github.com/subfuzion/db7f57fff2fb6998a16c for example

petergphillips (approved):
LGTM

.gitignore
tomsmyers: thanks for the tip, i actually didn't know this!

PR #530, opened 2020-09-09
DT-1108: 🔥 Remove username validation for client credentials

petergphillips:


Mjwillis (approved):
LGTM - lets see if anything breaks!

PR #532, opened 2020-09-10
Revert "take delius username from /users/{username}/details response instead of from login form input"

petergphillips:
Reverts ministryofjustice/hmpps-auth#525


SimonMitchellMOJ (approved):
LGTM

PR #536, opened 2020-09-10
DT-1110: 🗃🐛 Add unique constraint on user_token to prevent duplicates

petergphillips:
We're hydrating the user_token table into a Set so duplicates are automatically removed, so when we want to delete all the rows that doesn't work as the ones that were removed aren't processed so we get a foreign key violation.

### DIFF
--- a/src/main/java/uk/gov/justice/digital/hmpps/oauth2server/auth/model/User.java
+++ b/src/main/java/uk/gov/justice/digital/hmpps/oauth2server/auth/model/User.java
@@ -38,6 +38,7 @@ import java.util.UUID;
 
 import static javax.persistence.FetchType.EAGER;
 
+@SuppressWarnings("JpaDataSourceORMInspection")
 @Entity
 @Table(name = "USERS")
 @NoArgsConstructor
@@ -170,10 +171,15 @@ public class User implements UserPersonDetails, CredentialsContainer {
     }
 
     public UserToken createToken(final TokenType tokenType) {
-        final var token = new UserToken(tokenType, this);
-        // equals and hashcode by token type so remove will remove any token of same type
-        tokens.remove(token);
-        tokens.add(token);
+        final var optionalToken = tokens.stream().filter(t -> t.getTokenType() == tokenType).findFirst();
+        final UserToken token;
+        if (optionalToken.isPresent()) {
+            token = optionalToken.get();
+            token.resetExpiry();
+        } else {
+            token = new UserToken(tokenType, this);
+            tokens.add(token);
+        }
         return token;
     }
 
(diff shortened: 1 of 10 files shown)


src/main/resources/db/auth/V4_10__user_token_unique.sql
```diff
@@ -0,0 +1,7 @@
+delete
```
petergphillips, line 1: Remove duplicates so can add unique key.  Note that we remove all tokens for the user rather than just the token type, couldn't work out the correct SQL, but it will do.

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/auth/model/User.java
```diff
@@ -170,10 +171,15 @@ public String getUserId() {
     }
 
     public UserToken createToken(final TokenType tokenType) {
-        final var token = new UserToken(tokenType, this);
-        // equals and hashcode by token type so remove will remove any token of same type
-        tokens.remove(token);
```
petergphillips, line 175: So we were just removing the old token and then adding a new created one.  However hibernate adds the new record *before* deleting the old one, which breaks the unique constraint on the table.  Don't know a better way round it, so end up modifying the expiry of the existing one instead

src/test/java/uk/gov/justice/digital/hmpps/oauth2server/verify/ResetPasswordServiceTest.kt
```diff
@@ -135,7 +135,7 @@ class ResetPasswordServiceTest {
     whenever(userService.findMasterUserPersonDetails(anyString())).thenReturn(staffUserAccountForBobOptional)
     val existingUserToken = user.createToken(UserToken.TokenType.RESET)
     resetPasswordService.requestResetPassword("user", "url")
-    assertThat(user.tokens).hasSize(1).extracting<String> { obj: UserToken -> obj.token }.doesNotContain(existingUserToken.token)
+    assertThat(user.tokens).hasSize(1).extracting<String> { it.token }.containsExactly(existingUserToken.token)
```
petergphillips, line 138: since we're now replacing the existing token (and updating the expiry) the entity will be the same

src/test/java/uk/gov/justice/digital/hmpps/oauth2server/auth/repository/UserTokenRepositoryTest.kt
```diff
@@ -57,6 +64,20 @@ class UserTokenRepositoryTest {
     assertThat(retrievedEntity.user.username).isEqualTo("LOCKED_USER")
   }
 
+  @Test
+  fun `replacing existing token doesn't cause unique key violations`() {
```
petergphillips, line 68: bit naff that it was the integration tests that spotted the unique constraint issue, so have added a repository test to spot earlier if reoccurs

SimonMitchellMOJ (approved):
LGTM

PR #542, opened 2020-09-22
DT-1140: ♻️ Convert code to kotlin

petergphillips:


src/main/java/uk/gov/justice/digital/hmpps/oauth2server/auth/model/Authority.java
```diff
@@ -43,4 +47,16 @@ public static String removeRolePrefixIfNecessary(final String role) {
     public String getAuthority() {
         return ROLE_PREFIX + roleCode;
     }
+
+    public UUID getId() {
```
petergphillips, line 51: getters needed temporarily as kotlin compilation happens before lombok

Mjwillis (approved):
LGTM

PR #546, opened 2020-09-23
Add Azure OIDC User model and service

electrified:
This is the first PR in a set to make HMPPS Auth an Azure AD OIDC consumer.

Once a user has authenticated via Azure OpenID Connect, they will be stored in the auth DB with the Azure OID GUID as their username.

This change once all parts of the set are merged will enable users to sign in with their justice.gov.uk credentials. (When connected to the production Azure AD instance).

The original PR with the complete change is here: https://github.com/ministryofjustice/hmpps-auth/pull/520


src/main/java/uk/gov/justice/digital/hmpps/oauth2server/auth/repository/UserRepository.java
```diff
@@ -8,8 +8,9 @@
 import java.time.LocalDateTime;
 import java.util.List;
 import java.util.Optional;
+import java.util.UUID;
 
-public interface UserRepository extends CrudRepository<User, String>, JpaSpecificationExecutor<User> {
+public interface UserRepository extends CrudRepository<User, UUID>, JpaSpecificationExecutor<User> {
```
electrified, line 13: Correcting the primary key data type.

src/main/resources/db/auth/V4_11__increase_username_length.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE user_retries ALTER COLUMN username TYPE varchar(37);
+ALTER TABLE users ALTER COLUMN username TYPE varchar(37);
```
electrified, line 2: Column width increased by 7 characters so UUIDs can be stored

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/azure/service/AzureUserService.kt
```diff
@@ -0,0 +1,38 @@
...
+import lombok.extern.slf4j.Slf4j
+import org.slf4j.Logger
+import org.slf4j.LoggerFactory
+import org.springframework.stereotype.Service
+import uk.gov.justice.digital.hmpps.oauth2server.auth.repository.UserRepository
+import uk.gov.justice.digital.hmpps.oauth2server.azure.AzureUserPersonDetails
+import uk.gov.justice.digital.hmpps.oauth2server.security.AuthSource
+import java.util.*
+
+@Slf4j
```
petergphillips, line 12: lombok annotation :grinning:

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/azure/service/AzureUserService.kt
```diff
@@ -0,0 +1,38 @@
...
+  companion object {
+    val log: Logger = LoggerFactory.getLogger(this::class.java)
+  }
+
+  fun getAzureUserByUsername(username: String?): Optional<AzureUserPersonDetails> {
+    val user = userRepository.findByUsernameAndSource(username, AuthSource.azuread)
+
+    if (user.isPresent) {
+      return Optional.of(AzureUserPersonDetails(
+          Collections.emptyList(),
```
petergphillips, line 25: kotlin is `mutableListOf`

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/azure/service/AzureUserService.kt
```diff
@@ -0,0 +1,38 @@
...
+class AzureUserService(private val userRepository: UserRepository) {
+
+  companion object {
+    val log: Logger = LoggerFactory.getLogger(this::class.java)
+  }
+
+  fun getAzureUserByUsername(username: String?): Optional<AzureUserPersonDetails> {
+    val user = userRepository.findByUsernameAndSource(username, AuthSource.azuread)
+
+    if (user.isPresent) {
```
petergphillips, line 23: since the repository returns an `Optional` anyway this method can be simplied to
```
  fun getAzureUserByUsername(username: String?): Optional<AzureUserPersonDetails> =
      userRepository.findByUsernameAndSource(username, AuthSource.azuread)
          .map {
            AzureUserPersonDetails(
                mutableListOf(),
                true,
                it.username,
                it.person.firstName,
                it.person.lastName,
                it.email,
                credentialsNonExpired = true,
                accountNonExpired = true,
                accountNonLocked = true)
          }
```

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/security/AzureUserDetailsService.kt
```diff
@@ -0,0 +1,21 @@
...
+import org.springframework.security.core.userdetails.UsernameNotFoundException
+import org.springframework.security.web.authentication.preauth.PreAuthenticatedAuthenticationToken
+import org.springframework.stereotype.Service
+import uk.gov.justice.digital.hmpps.oauth2server.azure.AzureUserPersonDetails
+import uk.gov.justice.digital.hmpps.oauth2server.azure.service.AzureUserService
+
+@Service("azureUserDetailsService")
+class AzureUserDetailsService(private val azureUserService: AzureUserService) :
+    UserDetailsService, AuthenticationUserDetailsService<PreAuthenticatedAuthenticationToken> {
+  override fun loadUserByUsername(username: String?): AzureUserPersonDetails {
```
petergphillips, line 15: don't need the `?` - see the delius version

src/main/resources/db/auth/V4_11__increase_username_length.sql
petergphillips: For db changes worth running this SQL against a sqlserver database.  I've just tried it and `TYPE` isn't recognised.  We really need instructions in the README for how to do it, will DM

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/azure/service/AzureUserService.kt
```diff
@@ -0,0 +1,38 @@
...
+@Slf4j
+@Service
+class AzureUserService(private val userRepository: UserRepository) {
+
+  companion object {
+    val log: Logger = LoggerFactory.getLogger(this::class.java)
+  }
+
+  fun getAzureUserByUsername(username: String?): Optional<AzureUserPersonDetails> {
+    val user = userRepository.findByUsernameAndSource(username, AuthSource.azuread)
```
petergphillips, line 21: why is this going to the auth database to read an azure user?

petergphillips (changes requested):
SQL needs amending for sqlserver

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/azure/service/AzureUserService.kt
petergphillips: just read the description in the PR.  Need to ponder...

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/azure/AzureUserPersonDetails.kt
```diff
@@ -0,0 +1,55 @@
...
+data class AzureUserPersonDetails(private val authorities: MutableCollection<GrantedAuthority>,
+                                  private val enabled: Boolean,
+                                  private val username: String,
+                                  private val firstName: String,
+                                  val surname: String,
+                                  val email: String,
+                                  private val credentialsNonExpired: Boolean,
+                                  private val accountNonExpired: Boolean,
+                                  private val accountNonLocked: Boolean) : UserPersonDetails {
+  override fun getAuthorities(): Collection<GrantedAuthority?> = authorities
```
mikehalmamoj, line 18: Not 100% on this, but can you just put the override on the constructor parameters so you don't have to override all of the getter functions?

electrified: Thanks for the tip, I will investigate

electrified: Hi Mike. I tried and couldn't get it to work. It looks like I'd need this https://youtrack.jetbrains.com/issue/KT-6653

petergphillips (approved):
LGTM

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/azure/AzureUserPersonDetails.kt
mikehalmamoj: Ah it's an interoperability problem - it would be ok if the inherited interface was in Kotlin but not Java.  OK thanks for looking, interesting issue.

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/azure/service/AzureUserService.kt
electrified: This is an interesting question!

At the moment we are are considering that to get to this point the user must have authenticated via Azure and their account was valid at that point, to be adequate.

For future enhancements we have considered:
- Calling the Microsoft Graph API, which provides an API for getting a user by ID: https://docs.microsoft.com/en-us/graph/api/user-get?view=graph-rest-1.0&tabs=http#example-1-standard-users-request. This requires additional client permissions, but our client has been granted them on the DEVL Azure instance by Matt White.
- Using the Azure issued token to call the "me" endpoint detailed on the same page (or call the userinfo endpoint again). This would require quite a bit of refactoring to get the original Azure token to be available at this point in the code, but doesn't need any extra client permissions.

PR #547, opened 2020-09-23
AC-311 - Add Azure OIDC as an authentication source

electrified:
2nd change of the Azure OIDC change set.

This adds the ability to authenticate with Azure OIDC, however no user visible change (i.e. link on the login page) will be present yet.

The Azure journey can be started by visiting http://localhost:8080/auth/oauth2/authorization/microsoft (if you override the values in application-dev-config-azure-ad-test.yml with real values)


build.gradle.kts
```diff
@@ -18,10 +18,12 @@ dependencies {
   implementation("org.springframework.security:spring-security-jwt:1.1.1.RELEASE")
   implementation("org.springframework.security.oauth:spring-security-oauth2:2.5.0.RELEASE")
   implementation("io.jsonwebtoken:jjwt:0.9.1")
-  implementation("com.nimbusds:nimbus-jose-jwt:9.0.1")
```
electrified, line 21: This overrides the older version brought in by `spring-boot-starter-oauth2-client` and has some missing methods resulting in runtime exceptions when using the oauth2 client. Not seen any adverse effects from removing

src/main/resources/application-dev.yml
```diff
@@ -1,3 +1,3 @@
 spring:
   profiles:
-    include: dev-config,auth-seed,nomis-seed
+    include: dev-config,auth-seed,nomis-seed,dev-config-azure-ad-test
```
petergphillips, line 3: Not sure this is what we want, won't this turn on azure ad for 8101 for all users that want the `dev` profile?
We originally went down that route with token verification and delius integration, but since auth is run as a docker image in lots of projects we ended up turning it on instead when required in tests etc.

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/config/AuthenticationManagerConfiguration.java
```diff
@@ -125,9 +149,42 @@ protected void configure(final HttpSecurity http) throws Exception {
                 .addFilterAfter(jwtCookieAuthenticationFilter, BasicAuthenticationFilter.class)
 
                 .requestCache().requestCache(cookieRequestCache);
+
+        if (clientRegistrationRepository.isPresent()) {
```
petergphillips, line 153: what is the client registration repository and how is it configured?

src/main/resources/application-dev.yml
electrified: Agreed, yes it doesn't make much sense by itself. 

This profile is only needed when running the integration tests which are to follow in a later PR, which run Wiremock on 8101.

I'll remove it from this PR.

petergphillips: Quite happy for the profile to be there, just don't think it should be in triggered by the `dev` profile.

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/config/AuthenticationManagerConfiguration.java
electrified: It's constructed by the spring boot auto configuration, if you have any OAuth2 clients configured. It encapsulates that config basically.

See:

https://github.com/spring-projects/spring-boot/blob/master/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/security/oauth2/client/servlet/OAuth2ClientRegistrationRepositoryConfiguration.java#L43

Really I wanted a way to only run that section of code on `@Conditional(ClientsConfiguredCondition.class)`, as I do with the bean further down, but couldn't figure out how to do that.

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/config/AuthenticationManagerConfiguration.java
```diff
@@ -161,6 +218,7 @@ protected void configure(final AuthenticationManagerBuilder auth) {
         auth.authenticationProvider(deliusAuthenticationProvider);
         auth.authenticationProvider(preAuthProvider(authUserDetailsService));
         auth.authenticationProvider(preAuthProvider(nomisUserDetailsService));
+        auth.authenticationProvider(preAuthProvider(azureUserDetailsService));
```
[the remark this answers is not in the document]
electrified: Yes, you are correct. It's not explicitly configured (Spring Boot auto configuration again) but the initial token exchange is checked by the `OidcAuthorizationCodeAuthenticationProvider` invoked when the user gets redirected back to `/login/oauth2/code/microsoft` with an auth code.

petergphillips (approved):
LGTM

PR #549, opened 2020-09-24
AC-311 - Fix Username column width change to run on SQL Server. Also …

electrified:
…replace invalid UUIDs in seed data with correct ones


petergphillips (approved):
LGTM

PR #551, opened 2020-09-24
AC-311 - Add links to login page for authenticating via OAuth2 clients

electrified:
If OAuth2 clients are configured, show login links on the login page.

We only have a use case for 1 auth provider (Azure AD OIDC), but the code was simplified by making it show a link for each provider configured.

image


petergphillips (approved):
LGTM

PR #552, opened 2020-09-24
Azure OIDC integration tests

electrified:
This adds some config so that the Azure integration can be tested against a Wiremock simulation of Azure OIDC.

Do we want the test to be enabled like this when run on CircleCI? I can see the value in having the test disabled, and testing with a more production like config of having no OIDC clients configured. Open to thoughts on this!


petergphillips (approved):
LGTM

PR #558, opened 2020-09-29
AC-311 Add Azure AD documentation

electrified:
Open to the config being in an example yaml file rather than directly in the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -110,3 +110,45 @@ the authorization header, which allows password, authorization_code and refresh 
 This will go to github to get the specified version of the front end toolkit.  Very noddy at present, so if the assets have changed at https://github.com/alphagov/govuk-frontend/tree/master/dist/assets/fonts for the specified version in question, then the list in the bash script will need updating to cope.
 
 It will sort out the css references to `/assets` as we run in a `/auth` context instead. 
+
+### Enabling Azure AD OIDC as an additional authentication provider
+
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -110,3 +110,45 @@ the authorization header, which allows password, authorization_code and refresh
...
+#### Creating an Azure app registration
+1. Navigate to the app registrations page https://portal.azure.com/#blade/Microsoft_AAD_IAM/ActiveDirectoryMenuBlade/RegisteredApps
+2. Create a new app registration. For "Supported account types", select: "Accounts in any organizational directory (Any Azure AD directory - Multitenant)". Set the redirect URL to `http://localhost:8080/auth/login/oauth2/code/microsoft`
+3. Save the registration. Note the Client ID and Directory ID.
+4. Navigate to the Certificates & Secrets tab, and add a new client secret, and note the secret value
+
+#### Configuring Auth
+Add the following configuration to HMPPS Auth, replacing the `<placeholders>` with real values from the Azure App registration
+```
+spring:
```
petergphillips, line 131: Could all this be moved into a profile file instead with properties for `<Directory (tenant) ID>` etc?  The instructions could then be to set the `application.microsoft.directory-id` properties etc. and run with the specified profile as well

electrified: That's a great idea, I'll amend.

petergphillips (approved):
LGTM

PR #559, opened 2020-09-30
AC-312 add AzureAD -> nDelius user mapping 

tomsmyers:
this change adds a "UserContextService" class which is called when adding user details to access tokens. for all existing HMPPS Auth clients and users, this will have no effect. the default path through `UserContextService.resolveUser` is simply `return loginUser`.

however, if a client has been granted the `delius` scope, an attempt is made to "map" the login user to a user representing an nDelius account. if the login user is already an nDelius user (i.e. the `auth_source` is "delius"), nothing happens. if the login user is either a nomis or auth user, an _error_ is logged (this is very much intended - if a client is requesting the "delius" scope for users which have logged in via either of these methods, we want alerting). if the user is an "azuread" user, a lookup is performed on community-api and the delius user account information is returned in the token instead of the azure account information. if more than one ndelius user account is found for the given azuread account, a _warning_ is logged, and the original azuread account information is returned. this behaviour will be improved upon in the future.


src/main/java/uk/gov/justice/digital/hmpps/oauth2server/delius/service/DeliusUserService.kt
```diff
@@ -16,7 +16,9 @@ import uk.gov.justice.digital.hmpps.oauth2server.delius.model.DeliusUserPersonDe
 import uk.gov.justice.digital.hmpps.oauth2server.delius.model.UserDetails
 import uk.gov.justice.digital.hmpps.oauth2server.delius.model.UserRole
 import uk.gov.justice.digital.hmpps.oauth2server.security.DeliusAuthenticationServiceException
-import java.util.*
+import java.util.Optional
+
+class DeliusUserList : MutableList<UserDetails> by ArrayList()
```
electrified, line 21: It is possible to remove the need for this. Have had a go in another branch: `delius-id-no-deliususerlist`

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/delius/service/DeliusUserService.kt
```diff
@@ -29,6 +31,43 @@ class DeliusUserService(@Qualifier("deliusApiRestTemplate") private val restTemp
     private val log = LoggerFactory.getLogger(this::class.java)
   }
 
+  fun getDeliusUserByEmail(email: String): DeliusUserPersonDetails? {
+    if (!deliusEnabled) {
+      log.debug("delius integration disabled; unable to proceed for user with email {}", email)
```
electrified, line 36: Comment as before about capitalisation of log messages for consistency with existing messages :D

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/service/UserContextService.kt
```diff
@@ -0,0 +1,46 @@
...
+import uk.gov.justice.digital.hmpps.oauth2server.security.AuthSource
+import uk.gov.justice.digital.hmpps.oauth2server.security.AuthSource.azuread
+import uk.gov.justice.digital.hmpps.oauth2server.security.AuthSource.delius
+import uk.gov.justice.digital.hmpps.oauth2server.security.AuthSource.fromNullableString
+import uk.gov.justice.digital.hmpps.oauth2server.security.UserPersonDetails
+
+class UserMappingException(message: String): Exception(message)
+
+@Service
+class UserContextService(
```
electrified, line 15: Wondering about the name of this service. Does `Context` convey a useful meaning?

Mapping? Something else?

tomsmyers: the nomenclature came out of this discussion 

https://github.com/ministryofjustice/hmpps-auth/pull/517#discussion_r469772117

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/config/JWTTokenEnhancer.java
```diff
@@ -37,32 +41,45 @@
...
     @Override
     public OAuth2AccessToken enhance(final OAuth2AccessToken accessToken, final OAuth2Authentication authentication) {
         final Map<String, Object> additionalInfo;
 
         if (authentication.isClientOnly()) {
             additionalInfo = addUserFromExternalId(authentication, accessToken);
         } else {
-            final var userAuthentication = authentication.getUserAuthentication();
-            final var userDetails = (UserPersonDetails) userAuthentication.getPrincipal();
-            final var userId = StringUtils.defaultString(userDetails.getUserId(), userAuthentication.getName());
```
petergphillips, line 49: little bit worried that we're amending existing behaviour here, not sure why the defensive code was put in in the first place though

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/config/JWTTokenEnhancer.java
```diff
@@ -37,32 +41,45 @@
...
             final var clientDetails = clientsDetailsService.loadClientByClientId(authentication.getOAuth2Request().getClientId());
+            final var userDetails = getUser(authentication);
+
             // note that DefaultUserAuthenticationConverter will automatically add user_name to the access token, so
             // removal of user_name will only affect the authorisation code response and not the access token field.
             additionalInfo = filterAdditionalInfo(
-                    Map.of(SUBJECT, userAuthentication.getName(),
+                    Map.of(SUBJECT, userDetails.getUsername(),
                             ADD_INFO_AUTH_SOURCE, StringUtils.defaultIfBlank(userDetails.getAuthSource(), "none"),
-                            ADD_INFO_USER_NAME, userAuthentication.getName(),
```
petergphillips, line 57: again changing existing behaviour here, but not sure why the original code was there...

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/delius/service/DeliusUserService.kt
```diff
@@ -29,6 +31,43 @@ class DeliusUserService(@Qualifier("deliusApiRestTemplate") private val restTemp
...
+      return null
+    }
+
+    return try {
+      val userDetailsList = restTemplate.getForObject("/users/search/email/{email}/details", DeliusUserList::class.java, email)
```
petergphillips, line 41: should really filter out disabled users here too

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/delius/service/DeliusUserService.kt
```diff
@@ -29,6 +31,43 @@ class DeliusUserService(@Qualifier("deliusApiRestTemplate") private val restTemp
...
+      when (userDetailsList.size) {
+        1 -> mapUserDetailsToDeliusUser(userDetailsList[0])
+        0 -> {
+          log.debug("user with email {} not found in delius", email)
+          null
+        }
+        else -> {
+          log.warn("multiple users found for email {}; unable to proceed", email)
```
petergphillips, line 49: should be `info` here, nothing is going wrong with the application

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/config/JWTTokenEnhancer.java
```diff
@@ -37,32 +41,45 @@
...
-                            ADD_INFO_USER_ID, userId,
-                            ADD_INFO_NAME, userDetails.getName()), clientDetails);
+                            ADD_INFO_USER_NAME, userDetails.getUsername(),
+                            ADD_INFO_USER_ID, userDetails.getUserId(),
+                            ADD_INFO_NAME, userDetails.getName()),
+                    clientDetails);
         }
 
         ((DefaultOAuth2AccessToken) accessToken).setAdditionalInformation(additionalInfo);
```
petergphillips, line 68: Don't you need to replace the authorities too with the authorities on mapped user details, otherwise they won't get any roles?

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/service/UserContextService.kt
```diff
@@ -0,0 +1,46 @@
...
+  fun resolveUser(loginUser: UserPersonDetails, scopes: Set<String>): UserPersonDetails {
+    if (scopes.contains(deliusScope)) {
+      return map(loginUser, fromNullableString(loginUser.authSource), delius) ?: loginUser
+    }
+
+    return loginUser
+  }
+
+  private fun map(user: UserPersonDetails, from: AuthSource, to: AuthSource): UserPersonDetails? = when (from) {
+    to -> null
```
petergphillips, line 34: this is a bit messy, if you returned the `user` here then you won't need the `?:` on the caller either

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/service/UserContextService.kt
```diff
@@ -0,0 +1,46 @@
...
+    azuread -> mapFromAzureAD(user, to)
+    else -> throw UserMappingException("auth source '${from}' not supported")
+  }
+
+  private fun mapFromAzureAD(user: UserPersonDetails, to: AuthSource): UserPersonDetails? = when (to) {
+    delius -> {
+      log.debug("mapping user context from azure -> delius")
+      deliusUserService.getDeliusUserByEmail(user.userId)
```
petergphillips, line 42: is this right?  The `userId` holds the email address not the `username`?

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/service/UserContextService.kt
```diff
@@ -0,0 +1,46 @@
...
+
+  private fun map(user: UserPersonDetails, from: AuthSource, to: AuthSource): UserPersonDetails? = when (from) {
+    to -> null
+    azuread -> mapFromAzureAD(user, to)
+    else -> throw UserMappingException("auth source '${from}' not supported")
+  }
+
+  private fun mapFromAzureAD(user: UserPersonDetails, to: AuthSource): UserPersonDetails? = when (to) {
+    delius -> {
+      log.debug("mapping user context from azure -> delius")
```
petergphillips, line 41: not sure this log statement adds any value since it doesn't contain the email address and this is the only mapping that is available

src/main/resources/db/dev/data/auth/V900_0__oauth_sample_data.sql
```diff
@@ -35,6 +35,7 @@ VALUES ('omicuser','1200','{"jwtFields":"-user_name"}','SYSTEM_READ_ONLY','passw
        ('token-verification-auth-api-client','3600', '{}','ROLE_AUTH_TOKEN_VERIFICATION','client_credentials', null,'{bcrypt}$2a$10$hQPvQMNfbh2vhjTjDWHpoeN.iRFRddJJug6qtBkWzQ8uPEc53isZy', null, null,'read', null),
        ('v1-client','1200','{}','ROLE_NOMIS_API_V1,ROLE_BOOKING_CREATE,ROLE_BOOKING_RECALL,ROLE_GLOBAL_SEARCH','client_credentials','read,write','$2a$10$r12DB/sqXduodnjtAY/ykO0S3KCySdVW4zhG3jlIRaIsfVkFOEds2',null,null,'read,write',null),
        ('probation-offender-search-indexer-client','3600','{}','ROLE_PROBATION_INDEX,ROLE_COMMUNITY','client_credentials',null,'$2a$10$lBwbziQlLfiCnn8Kj1PfMujEcLdsJYlYSNJvBRO638gCYTS9yN0xm',43200,null,'read',null),
+       ('delius-login-client','28800','{}','ROLE_ACCESS_DELIUS_ID,ROLE_COMMUNITY','authorization_code,refresh_token','read,delius','$2a$10$RYwV0QebHAovVXWPySb2lefr3HTDntGu1euXHDJc3zwh2NsqeNGHG','43200',null,'read,delius','http://localhost:5000/login/callback'),
```
petergphillips, line 38: Shouldn't be granting new clients `refresh_token` rights.  Also setting roles for `authorization_code` doesn't make any sense, they will be ignored - roles will come from the logged in user instead

src/test/java/uk/gov/justice/digital/hmpps/oauth2server/service/UserContextServiceTest.kt
```diff
@@ -0,0 +1,69 @@
...
+  private val deliusUserService: DeliusUserService = mock()
+  private val userContextService = UserContextService(deliusUserService)
+
+  @Test
+  fun `resolveUser returns the same user for clients with 'normal' scopes`() {
+    val loginUser = User.builder().username("username").source(auth).build()
+
+    var scopes = setOf("read")
+    var user = userContextService.resolveUser(loginUser, scopes)
+    assertEquals(user, loginUser)
```
petergphillips, line 23: we use assertJ rather than `assertEquals` etc in this project

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/service/UserContextService.kt
tomsmyers: yeh this looks strange, but it's correct

src/test/java/uk/gov/justice/digital/hmpps/oauth2server/service/UserContextServiceTest.kt
```diff
@@ -0,0 +1,69 @@
...
+    val user = userContextService.resolveUser(loginUser, scopes)
+    assertEquals(user, loginUser)
+  }
+
+  @Test
+  fun `resolveUser can map from azureAD to delius`() {
+    val deliusUser = DeliusUserPersonDetails("username", "id", "user", "name", "email@email.com")
+    val loginUser = UserDetailsImpl("username", "name", listOf(), "azuread", "email@email.com", "jwtId")
+    val scopes = setOf("delius")
+    whenever(deliusUserService.getDeliusUserByEmail("email@email.com")).thenReturn(deliusUser)
```
petergphillips, line 54: Better to be lenient on the mocking and then `verify` that the mock was called correctly, so `anyString` would be better here.  Stops hunting around to work out why tests are no longer going down the expected paths...

src/test/java/uk/gov/justice/digital/hmpps/oauth2server/delius/service/DeliusUserServiceTest.kt
```diff
@@ -22,6 +25,77 @@ class DeliusUserServiceTest {
   private val disabledDeliusService = DeliusUserService(restTemplate, false, mappings)
   private val deliusService = DeliusUserService(restTemplate, true, mappings)
 
+  @Test
+  fun `getDeliusUserByEmail disabled`() {
+    disabledDeliusService.getDeliusUserByEmail("a@where.com")
+    verify(restTemplate, never()).getForObject<UserDetails>(anyString(), any(), anyString())
```
petergphillips, line 31: would be better to `verifyZeroInteractions` with the `restTemplate`

src/test/java/uk/gov/justice/digital/hmpps/oauth2server/service/UserContextServiceTest.kt
```diff
@@ -0,0 +1,69 @@
+package uk.gov.justice.digital.hmpps.oauth2server.service
+
+import com.nhaarman.mockitokotlin2.*
```
petergphillips, line 3: prefer expanding out wildcard imports

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/config/JWTTokenEnhancer.java
tomsmyers: i think these methods are really the same thing

`org.springframework.security.core.userdetails.UserDetails` doesn't have a `name` field, only `username`. and `java.security.Principal` doens't have a `username` field, only `name`.

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/config/JWTTokenEnhancer.java
tomsmyers: i'll add this back in

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/service/UserContextService.kt
tomsmyers: done

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/config/JWTTokenEnhancer.java
tomsmyers: i'll address this at a later time 👍

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/delius/service/DeliusUserService.kt
tomsmyers: based on slack discussion, can we punt the decision on this to a later time? we'll talk it over with the others first.

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/delius/service/DeliusUserService.kt
tomsmyers: done

src/test/java/uk/gov/justice/digital/hmpps/oauth2server/delius/service/DeliusUserServiceTest.kt
tomsmyers: tbh i just copied this test from the equivalent for `getDeliusUserByUsername` below. best to keep it consistent with existing pattern in this file?

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/config/JWTTokenEnhancer.java
petergphillips: agreed, wasn't sure so raised :grinning:

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/delius/service/DeliusUserService.kt
tomsmyers: i think the only downside to that approach is that it requires instantiation of the list to extract the class at runtime, which doesn't seem ideal.

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/delius/service/DeliusUserService.kt
tomsmyers: done

petergphillips (approved):
LGTM

PR #561, opened 2020-10-01
Ac374 skip to azure config

rgforsyth:


src/main/resources/db/dev/data/auth/V900_0__oauth_sample_data.sql
```diff
@@ -39,11 +39,12 @@ VALUES ('omicuser','1200','{"jwtFields":"-user_name"}','SYSTEM_READ_ONLY','passw
        ('hmpps-ppm-client','3600','{}',null,'authorization_code','read,write','$2a$10$YRkR9FGSpZu3FAn5.Awtk.Yd0hg92y63VfVVAKhS6k66nMsc3/Hiy',43200,null,'read,write','http://localhost:3000/login/callback,http://localhost:3000'),
        ('hmpps-ppm-system','3600','{}','ROLE_GLOBAL_SEARCH','client_credentials','read','$2a$10$YRkR9FGSpZu3FAn5.Awtk.Yd0hg92y63VfVVAKhS6k66nMsc3/Hiy',43200,null,'read',null),
        ('manage-intelligence-client','3600','{}',null,'authorization_code','read,write','$2a$10$WzgtydqXSuhdivpWDR3WXO.yjLBm4yuDqP64Og.7E4XURdrSfhOTi',43200,null,'read,write','http://localhost:3000/login/callback,http://localhost:3000'),
-       ('manage-intelligence-admin','3600','{}','ROLE_GLOBAL_SEARCH,ROLE_SYSTEM_USER,ROLE_COMMUNITY','client_credentials','read,write','$2a$10$ajGimbJNWF1/FmZQMJWvieeQ/OdYaxWHQPgOjYDvvWu/4/744Yw7S',43200,null,'read,write',null);
+       ('manage-intelligence-admin','3600','{}','ROLE_GLOBAL_SEARCH,ROLE_SYSTEM_USER,ROLE_COMMUNITY','client_credentials','read,write','$2a$10$ajGimbJNWF1/FmZQMJWvieeQ/OdYaxWHQPgOjYDvvWu/4/744Yw7S',43200,null,'read,write',null),
```
rgforsyth, line 42: Will be removed after the AC312 branch is merged

src/main/resources/db/dev/data/auth/V900_0__oauth_sample_data.sql
```diff
@@ -39,11 +39,12 @@ VALUES ('omicuser','1200','{"jwtFields":"-user_name"}','SYSTEM_READ_ONLY','passw
...
+       ('skip-to-azure-client','3600','{}','ROLE_GLOBAL_SEARCH,ROLE_SYSTEM_USER,ROLE_COMMUNITY','authorization_code','read,write','$2a$10$RYwV0QebHAovVXWPySb2lefr3HTDntGu1euXHDJc3zwh2NsqeNGHG',43200,null,'read,write','http://localhost:5000/login/callback');
```
rgforsyth, line 43: Will be removed after the AC312 branch is merged

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/resource/ClientsController.java
```diff
@@ -103,5 +103,13 @@ public String getJwtFields() {
         public void setJwtFields(final String jwtFields) {
             addAdditionalInformation("jwtFields", jwtFields);
         }
+
+        public Boolean getSkipToAzureField(){
```
electrified, line 107: (nitpick) Formatting - could do with a space between ) and {, and on the method below

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/resource/LoginController.java
```diff
@@ -22,16 +26,44 @@
 @Controller
 public class LoginController {
     final private List<ClientRegistration> clientRegistrations;
+    private final CookieRequestCache cookieRequestCache;
+    
+    @Autowired
+    private ClientDetailsService clientDetailsService;
```
electrified, line 32: For consistency this should be autowired via the constructor as with the cookieRequestCache and clientRegistrations

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/resource/LoginController.java
```diff
@@ -22,16 +26,44 @@
...
+
+            final var isOAuthLogin = redirectUrl.getPath().endsWith("/oauth/authorize") && clientId != null;
+
+            if(isOAuthLogin) {
+                final var clientDetails = clientDetailsService.loadClientByClientId(clientId);
+
+                final Boolean skipToAzure = (Boolean) clientDetails.getAdditionalInformation().getOrDefault("skipToAzureField", false);
+
+                if (skipToAzure) {
+                    return new ModelAndView("redirect:/oauth2/authorization/microsoft");
```
electrified, line 61: It would be good to not hard code this (in case someone has set up the OAuth2 client with a different name) - the redirect URL will be available in the clientRegistrations.

I guess pick the first one from the list as there should only be 1 OAuth2 configuration for the time being?

src/main/resources/db/dev/data/auth/V900_0__oauth_sample_data.sql
electrified, line 43: Should the additional_info field have the skipTOAzure flag set on in it?

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/resource/LoginController.java
```diff
@@ -22,16 +26,44 @@
...
                 .orElse(Collections.emptyList());
+        this.cookieRequestCache = cookieRequestCache;
     }
 
     @GetMapping("/login")
-    public ModelAndView loginPage(@RequestParam(required = false) final String error) {
+    public ModelAndView loginPage(@RequestParam(required = false) final String error,
+                                  final HttpServletRequest request, final HttpServletResponse response) {
+
+        final var savedRequest = cookieRequestCache.getRequest(request, response);
```
electrified, line 47: Before attempting any of the redirect logic, test that there is at least 1 OAuth2 client configured. (clientRegistrations has length >= 1)

src/test/java/uk/gov/justice/digital/hmpps/oauth2server/resource/LoginControllerTest.kt
```diff
@@ -31,22 +57,67 @@ class LoginControllerTest {
...
+    var returnRequest = SimpleSavedRequest("test.com/test?client_id=test_id")
+
+    whenever(cookieRequestCacheMock.getRequest(ArgumentMatchers.any(), ArgumentMatchers.any())).
+    thenReturn(returnRequest)
+    val modelAndView = controller.loginPage(null, null, null)
+
+    assertThat(modelAndView.viewName).isEqualTo("login")
+   }
+
+
```
electrified, line 121: Tests look good!

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/resource/LoginController.java
electrified, line 47: Before doing any of the redirect logic, check that there is an enabled OAuth2 client to redirect to (clientRegistrations length >= 1)

src/main/resources/db/dev/data/auth/V900_0__oauth_sample_data.sql
electrified, line 43: Should the additional_info have the skip to azure flag set?

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/resource/LoginController.java
```diff
@@ -22,16 +26,44 @@
...
+
+        if (savedRequest != null) {
+            final var redirectUrl = UriComponentsBuilder.fromUriString(savedRequest.getRedirectUrl()).build();
```
electrified, line 50: Wondering if it would be worth extracting some of this out into a separate method? Just an idea - not sure if it's worthwhile

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/resource/ClientsController.java
petergphillips: wondering if this is more of a `skipToClientRegistration` instead?

petergphillips: or `redirectLoginToClientRegistration`

src/main/resources/templates/ui/form.html
```diff
@@ -112,6 +112,19 @@ <h1 th:if="${param.client}" class="govuk-heading-l">Edit client '<span th:text="
...
+      <label class="govuk-label" for="skipToAzureField">Azure Ad Login Flow</label>
+      <span class="govuk-hint">
+          Customize behaviour of Azure AD login (BETA)
+        </span>
+      <div class="govuk-checkboxes__item">
+        <input name="_skipToAzureField" type="hidden" value="on" />
+        <input class="govuk-checkboxes__input" type="checkbox" name="skipToAzureField" id="skipToAzureField"
+               th:checked="*{clientDetails.additionalInformation['skipToAzureField']}"/>
+        <label class="govuk-label govuk-checkboxes__label" for="skipToAzureField">Auto redirect</label>
+      </div>
```
electrified, line 125: Should we make it so you can specify which client you get redirected to? Or is that overkill as we really only ever forsee there being one.

src/test/java/uk/gov/justice/digital/hmpps/oauth2server/resource/LoginControllerTest.kt
```diff
@@ -1,28 +1,54 @@
 package uk.gov.justice.digital.hmpps.oauth2server.resource
 
+import com.nhaarman.mockitokotlin2.any
+import com.nhaarman.mockitokotlin2.mock
+import com.nhaarman.mockitokotlin2.whenever
 import org.assertj.core.api.Assertions.assertThat
+import org.junit.Before
```
petergphillips, line 7: odd unused imports.
Normally in intellij we click the boxes to reformat and remove unused imports when committing, which will sort the issues with the formatting etc. in this PR.

src/main/resources/db/dev/data/auth/V900_0__oauth_sample_data.sql
rgforsyth: It can. If it is blank or set to false, it will be disabled. If set too true {"skipToAzure" : true} then it would be active, and the checkbox checked upon loading. After the form has been saved it will be there regardless,  in the additional information, set to true/false

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/resource/LoginController.java
rgforsyth: Would this then negate the need for the first check later: 
`isOAuthLogin = redirectUrl.getPath().endsWith("/oauth/authorize") && clientId != null;`

tomsmyers:
this all looks good to me. regarding configuring which client you get ridirected to etc, my feeling is keep it simple and keep this feature focused specifically on azure AD. the only annoying thing about that is that is that the azure ad client is quite generic itself, just being `clientRegistrations[0]` or whatever. so it doesn't look like it's specifically for azure ad. i guess you could just document this in the code by extracting to a var called `azureADClient` or something. e.g. 

```kotlin
val azureADClient = clientRegistrations.getOrNull(0)
if azureADClient == null return
```

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/resource/ClientsController.java
rgforsyth: It could be, however I think the plan is to keep this change specifically focused on the skipping to 'AzureAD' as Tom mentioned below.

src/main/resources/templates/ui/form.html
rgforsyth: I think we decided to make the changes specific too skipping to azure for this change.

src/test/java/uk/gov/justice/digital/hmpps/oauth2server/resource/LoginControllerTest.kt
```diff
@@ -1,28 +1,46 @@
...
   private val clientRegistrationRepository: Optional<InMemoryClientRegistrationRepository> = Optional.empty()
+  private val cookieRequestCacheMock: CookieRequestCache = mock()
+  private val clientDetailsService: ClientDetailsService = mock()
+
+  private val controller = LoginController(clientRegistrationRepository, cookieRequestCacheMock, clientDetailsService)
+
+  @BeforeEach
+  internal fun setUp() {
+    ReflectionTestUtils.setField(controller, "clientDetailsService", clientDetailsService)
+  }
```
electrified, line 31: Now you have added clientDetailsService as a constructor parameter, there is no need to use reflection to populate it.

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/resource/LoginController.java
```diff
@@ -22,16 +25,42 @@
 @Controller
 public class LoginController {
     final private List<ClientRegistration> clientRegistrations;
+    private final CookieRequestCache cookieRequestCache;
+    private ClientDetailsService clientDetailsService;
```
electrified, line 29: Think this one can be `private final` too?

electrified (approved):
LGTM

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/resource/LoginController.java
```diff
@@ -22,16 +24,42 @@
...
+        this.clientDetailsService = clientDetailsService;
     }
 
     @GetMapping("/login")
-    public ModelAndView loginPage(@RequestParam(required = false) final String error) {
+    public ModelAndView loginPage(@RequestParam(required = false) final String error,
+                                  final HttpServletRequest request, final HttpServletResponse response) {
+
+        final var savedRequest = cookieRequestCache.getRequest(request, response);
+        if (savedRequest != null && clientRegistrations.size() >= 1) {
```
petergphillips, line 46: would be better with an `!isEmpty()` instead of checking the `size`

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/resource/LoginController.java
```diff
@@ -22,16 +24,42 @@
...
+            final var redirectUrl = UriComponentsBuilder.fromUriString(savedRequest.getRedirectUrl()).build();
+            final String clientId = redirectUrl.getQueryParams().getFirst("client_id");
+            final var isOAuthLogin = redirectUrl.getPath().endsWith("/oauth/authorize") && clientId != null;
+
+            if(isOAuthLogin) {
+                final var clientDetails = clientDetailsService.loadClientByClientId(clientId);
+                final Boolean skipToAzure = (Boolean) clientDetails.getAdditionalInformation().getOrDefault("skipToAzureField", false);
+
+                if (skipToAzure) {
+                    var azureADClient = clientRegistrations.get(0).getClientName();
```
petergphillips, line 56: inconsistency with missing `final` here, but `var`.  Don't worry about fixing it, there have been quite a few PRs with it missing recently.  I'll reformat and fix the code base afterwards.

petergphillips (approved):
LGTM

PR #566, opened 2020-10-02
DT-1172 Make Userinfo endpoint return same userid value as token for Azure OIDC users

electrified:


src/main/java/uk/gov/justice/digital/hmpps/oauth2server/azure/AzureUserPersonDetails.kt
```diff
@@ -29,7 +29,7 @@ data class AzureUserPersonDetails(private val authorities: MutableCollection<Gra
 
   override fun isAccountNonLocked(): Boolean = accountNonLocked
 
-  override fun getUserId(): String = username
+  override fun getUserId(): String = email
```
petergphillips, line 32: looks fine, but would be good to have a test as this is an add configuration to set

petergphillips (approved):
LGTM, although test would be good

PR #567, opened 2020-10-02
DT-1172: 🔧 Add in azure oidc config for dev (t3)

petergphillips:
Will add a login with a justice.gov.uk link to t3.  This only allows you to login at present and doesn't assign roles etc.  Furthermore confusingly digital.justice.gov.uk accounts don't count as aren't hosted in azure ad.

Screenshot from 2020-10-02 16-46-27


electrified (approved):
LGTM

PR #568, opened 2020-10-05
DT-1179: ✨ Add mapping from azure to external users

petergphillips:


src/main/java/uk/gov/justice/digital/hmpps/oauth2server/delius/service/DeliusUserService.kt
```diff
@@ -31,25 +32,15 @@ class DeliusUserService(@Qualifier("deliusApiRestTemplate") private val restTemp
     private val log = LoggerFactory.getLogger(this::class.java)
   }
 
-  fun getDeliusUserByEmail(email: String): DeliusUserPersonDetails? {
+  fun getDeliusUsersByEmail(email: String): List<DeliusUserPersonDetails> {
```
petergphillips, line 35: we effectively return all the matching users and let the caller determine if that's okay

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/service/UserContextService.kt
```diff
@@ -1,46 +1,48 @@
...
     }
-
-    return loginUser
   }
 
-  private fun map(user: UserPersonDetails, from: AuthSource, to: AuthSource): UserPersonDetails = when (from) {
-    to -> user
-    azuread -> mapFromAzureAD(user, to)
-    else -> throw UserMappingException("Auth source '${from}' not supported")
+  fun discoverUsers(loginUser: UserPersonDetails, scopes: Set<String>): List<UserPersonDetails> {
```
petergphillips, line 30: this method (currently unused outside of this class) can be used by the pre approval flow to see how many we have matching the `scopes`, although might be better to eventually be a `boolean`

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/service/UserContextService.kt
```diff
@@ -1,46 +1,48 @@
...
+    val loginUserAuthSource = fromNullableString(loginUser.authSource)
+    if (loginUserAuthSource != azuread) return listOf(loginUser)
```
petergphillips, line 32: deliberately only allow azure ad users past this point

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/service/UserContextService.kt
```diff
@@ -1,46 +1,48 @@
...
 
-  private fun mapFromAzureAD(user: UserPersonDetails, to: AuthSource): UserPersonDetails = when (to) {
-    delius -> {
-      log.debug("Mapping user context from azure -> delius")
-      deliusUserService.getDeliusUserByEmail(user.userId) ?: user
-    }
-    else -> throw UserMappingException("Auth -> '${to}' mapping not supported")
+  private fun mapFromAzureAD(user: UserPersonDetails, to: AuthSource): List<UserPersonDetails> = when (to) {
+    delius -> deliusUserService.getDeliusUsersByEmail(user.userId)
+    auth -> authUserService.findAuthUsersByEmail(user.userId)
```
petergphillips, line 45: new `auth` mapping of users

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/delius/service/DeliusUserService.kt
andymarke: I can't work out what actually happens if multiple Delius accounts are found for the email address? Is there an extra page in the login process or something like that?

andymarke (approved):
LGTM

PR #570, opened 2020-10-06
PAT-1346 Enforce MFA for all Pathfinder users

steverendell:
This is the 'quick' solution. Hopefully itll be followed up later by a solution controlled by the app rather than roles


petergphillips (approved):
LGTM

PR #578, opened 2020-10-08
DT-1191: 🎨 Enforce formatting of kotlin files by the build

petergphillips:


build.gradle.kts
```diff
@@ -99,3 +100,7 @@ tasks {
     setMaxHeapSize("256m")
   }
 }
+
+tasks.named("check") {
+  dependsOn(":ktlintCheck")
```
petergphillips, line 105: this means that the build will automatically run the check and running `test` etc from intellij / commmand line will too

SimonMitchellMOJ (approved):
LGTM

PR #583, opened 2020-10-09
DT-1072: ✨ Map multi user and single user azuread

petergphillips:
At present will just work for the single user matched scenario.
Need to add some tests for the changes.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,9 +1,12 @@
-[*.{kt,kts}]
+[*.{kt, kts}]
 indent_size = 2
-max_line_length = 120 
+max_line_length = 120
 
 [*.java]
 ij_java_class_count_to_use_import_on_demand = 99
 ij_java_names_count_to_use_import_on_demand = 99
 ij_java_use_single_class_imports = true
 ij_java_insert_inner_class_imports = true
+
+[*.html]
+indent_size = 2
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,6 +1,10 @@
-[*.{kt, kts}]
+[*.kt]
 indent_size = 2
-max_line_length = 120
+max_line_length = none
+
+[*.kts]
+indent_size = 2
+max_line_length = none
 
 [*.java]
 ij_java_class_count_to_use_import_on_demand = 99
--- a/src/main/java/uk/gov/justice/digital/hmpps/oauth2server/config/JWTTokenEnhancer.java
+++ b/src/main/java/uk/gov/justice/digital/hmpps/oauth2server/config/JWTTokenEnhancer.java
@@ -1,7 +1,6 @@
 package uk.gov.justice.digital.hmpps.oauth2server.config;
 
 
-import lombok.extern.slf4j.Slf4j;
 import org.apache.commons.lang3.StringUtils;
 import org.jetbrains.annotations.NotNull;
 import org.springframework.beans.factory.annotation.Autowired;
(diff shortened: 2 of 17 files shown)


src/test/java/uk/gov/justice/digital/hmpps/oauth2server/integration/ClientLoginSpecification.kt
```diff
@@ -122,6 +122,12 @@ class ClientLoginSpecification : AbstractAuthSpecification() {
       .jsonPath(".user_id").isEqualTo("4")
       .jsonPath(".sub").isEqualTo("RO_USER")
       .jsonPath(".auth_source").isEqualTo("nomis")
+      .jsonPath(".access_token").value<JSONArray> {
+        val claims = JWTParser.parse(it[0].toString()).jwtClaimsSet
+        assertThat(claims.getClaim("user_name")).isEqualTo("RO_USER")
+        assertThat(claims.getClaim("authorities") as List<*>)
+          .containsExactly("ROLE_GLOBAL_SEARCH", "ROLE_PRISON", "ROLE_INACTIVE_BOOKINGS", "ROLE_LICENCE_RO")
```
petergphillips, line 129: we get the mapping of roles for free for the single user case with the new approach

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/config/JWTTokenEnhancer.java
```diff
@@ -41,45 +37,32 @@
     @Autowired
     private JdbcClientDetailsService clientsDetailsService;
 
-    @Autowired
-    private UserContextService userContextService;
```
petergphillips, line 45: revert all these changes, will be done earlier in the chain by the authorization endpoint instead

.editorconfig
```diff
@@ -1,9 +1,16 @@
-[*.{kt,kts}]
```
petergphillips, line 1: seemed to be struggling with this line, kept on reformatting the build file with 4 spaces instead.  Splitting out fixed it.

.editorconfig
```diff
@@ -1,9 +1,16 @@
...
+[*.kt]
 indent_size = 2
-max_line_length = 120 
+max_line_length = none
```
petergphillips, line 3: we weren't actually using 120, so corrected

.editorconfig
```diff
@@ -1,9 +1,16 @@
...
 
 [*.java]
 ij_java_class_count_to_use_import_on_demand = 99
 ij_java_names_count_to_use_import_on_demand = 99
 ij_java_use_single_class_imports = true
 ij_java_insert_inner_class_imports = true
+
+[*.html]
+indent_size = 2
```
petergphillips, line 16: sort out web page format too

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/resource/UserSelectorAuthorizationEndpoint.kt
```diff
@@ -0,0 +1,122 @@
...
+
+    val replacedAuthentication = if (!account.isNullOrBlank() && authentication != null) {
+      val source = account.substringBefore("/")
+      val username = account.substringAfter("/")
+
+      val azureUser = authentication.principal as UserDetailsImpl
+
+      val user = userService.getMasterUserPersonDetails(username, AuthSource.fromNullableString(source))
+      user.map {
+        // TODO: ensure that the email address matches for the user
```
petergphillips, line 98: will do this in next PR

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/service/UserContextService.kt
```diff
@@ -32,7 +32,7 @@ class UserContextService(
     }
   }
 
-  private fun discoverUsers(loginUser: UserPersonDetails, scopes: Set<String>): List<UserPersonDetails> {
```
petergphillips, line 35: will remove the unnecessary `resolveUser` in next PR as using new method here instead

.editorconfig
mikehalmamoj: I also vote for `ij_kotlin_call_parameters_wrap = off`.

JamieDraperUK (approved):
LGTM!

PR #586, opened 2020-10-13
DT-1201: 🎨 Sort out formatting from previous PR

petergphillips:


.editorconfig
```diff
@@ -1,10 +1,8 @@
 [*.kt]
 indent_size = 2
-max_line_length = none
```
petergphillips, line 3: default is `off` anyway so wasn't doing anything

SimonMitchellMOJ (approved):
LGTM

PR #597, opened 2020-10-20
DT-1165: 🗃 Modify PECS groups

petergphillips:


src/main/resources/db/auth/V4_14__pecs_group_additions.sql
```diff
@@ -0,0 +1,45 @@
...
+
+INSERT INTO groups (group_id, group_code, group_name)
+VALUES (NEWID(), 'PECS_GMP13', 'PECS Manchester Airport Police Station');
+
+INSERT INTO group_assignable_role (role_id, group_id, automatic)
+VALUES ((select role_id from roles where role_code = 'PECS_POLICE'),
+        (select group_id from groups where group_code in ('PECS_GMP13')),
+        1);
+
+INSERT INTO user_group (group_id, user_id)
```
petergphillips, line 33: note only the police group needs users copying over.  Chosen GMP8 as there is no GMP12

JamieDraperUK (approved):
lgtm

PR #600, opened 2020-10-20
DT 1225 fetch and save nomis email on first login

SimonMitchellMOJ:


src/main/java/uk/gov/justice/digital/hmpps/oauth2server/security/UserRetriesService.kt
```diff
@@ -26,7 +29,9 @@ class UserRetriesService(
     resetRetries(username)
     // and record last logged in as now too (doing for all users to prevent confusion)
     val userOptional = userRepository.findByUsername(username)
-    val user = userOptional.orElseGet { userPersonDetails.toUser() }
+    val user = userOptional.orElseGet {
+      (userPersonDetails as? NomisUserPersonDetails)?.let { addNomisEmail(it, username) } ?: userPersonDetails.toUser()
```
petergphillips, line 33: haven't seen the `as?` before, looks good

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/security/UserRetriesService.kt
```diff
@@ -36,6 +41,15 @@ class UserRetriesService(
     userRepository.save(user)
   }
 
+  private fun addNomisEmail(userPersonDetails: UserPersonDetails, username: String): User? {
```
petergphillips, line 44: can be `User` since never returning `null`

petergphillips (approved):
LGTM

PR #601, opened 2020-10-20
DT-1213: Update user-load to send all groups in initial user creation

electrified:


user-load.sh
```diff
@@ -94,9 +83,15 @@ while IFS=, read -r -a row; do
 
   echo "Processing ${row[*]}"
 
+  groups=""
+
+  for group in "${row[@]:4}"; do
+    groups+="\"${group}\","
```
petergphillips, line 89: this can be simplified to:
```
  printf -v groups '\"%s\",' "${row[@]:4}"
```
which then adds a trailing comma at the end which can be removed in the `curl` line with `[${groups%,}]` instead.
Obviously not my code, grabbed from https://stackoverflow.com/questions/13470413/converting-a-bash-array-into-a-delimited-string

petergphillips (approved):
LGTM

PR #602, opened 2020-10-21

petergphillips:


src/main/java/uk/gov/justice/digital/hmpps/oauth2server/config/AuthDbConfig.kt
```diff
@@ -0,0 +1,51 @@
...
+@Configuration
+@EnableTransactionManagement
+@EnableJpaRepositories(
+  entityManagerFactoryRef = "authEntityManagerFactory",
+  transactionManagerRef = "authTransactionManager",
+  basePackages = ["uk.gov.justice.digital.hmpps.oauth2server.auth.repository"]
+)
+@ConfigurationProperties(prefix = "auth")
+class AuthDbConfig {
+  lateinit var jpa: Map<String, String>
```
petergphillips, line 26: tried playing around with constructor injection but couldn't get it to work as is.  Could get it to work with @ConfigurationPropertiesBinding and a separate data class but didn't want that.

src/test/java/uk/gov/justice/digital/hmpps/oauth2server/resource/LoginControllerTest.kt
```diff
@@ -57,7 +62,7 @@ class LoginControllerTest {
 
     val controller = LoginController(clientRegistrationRepository, cookieRequestCacheMock, clientDetailsService)
 
-    val modelAndView = controller.loginPage(null, null, null)
+    val modelAndView = controller.loginPage(null, request, response)
```
petergphillips, line 65: had to change these as they were passing through to the cookie helper as `null` which wasn't allowed after conversion.  They will never be `null` with the app running.

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/config/AuthenticationManagerConfiguration.kt
```diff
@@ -0,0 +1,214 @@
+@file:Suppress("DEPRECATION")
```
andymarke, line 1: Is this still needed?

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/config/AuthenticationManagerConfiguration.kt
```diff
@@ -0,0 +1,214 @@
...
+  private val deliusAuthenticationProvider: DeliusAuthenticationProvider,
+  private val userStateAuthenticationFailureHandler: UserStateAuthenticationFailureHandler,
+  private val clearAllSessionsLogoutHandler: ClearAllSessionsLogoutHandler,
+  private val clientRegistrationRepository: Optional<InMemoryClientRegistrationRepository>,
+) : WebSecurityConfigurerAdapter() {
+
+  override fun configure(http: HttpSecurity) {
+
+    // @formatter:off
+    http
```
andymarke, line 78: it would be interesting to look at the Kotlin Spring security DSL for this at some point

andymarke (approved):
LGTM

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/config/AuthenticationManagerConfiguration.kt
petergphillips: yes, since we are using spring security oauth that is still deprecated

PR #608, opened 2020-10-21

petergphillips:


src/main/java/uk/gov/justice/digital/hmpps/oauth2server/resource/ResetPasswordController.kt
```diff
@@ -62,7 +62,7 @@ class ResetPasswordController(
     }
     if (usernameOrEmail.contains("@")) {
       try {
-        verifyEmailService.validateEmailAddressExcludingGsi(EmailHelper.format(usernameOrEmail), User.EmailType.PRIMARY)
+        verifyEmailService.validateEmailAddressExcludingGsi(EmailHelper.format(usernameOrEmail)!!, User.EmailType.PRIMARY)
```
petergphillips, line 65: this is a bit naff, but some of the other calls to the helper can end up with a `null`, but not this one.  Hence helper has to still return a `String?`

mikehalmamoj: Better to have 2 methods - one with a nullable parameter and one with non-nullable parameter?

mikehalmamoj (approved):

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/resource/ResetPasswordController.kt
petergphillips: agreed, but didn't want to rewrite everything on the conversion

PR #609, opened 2020-10-21
DT-1242 DT-1243 remove unverified auth users from selection screen / only show selection screen when nessecary

tomsmyers:
this PR fixes two bugs and includes some refactoring in `UserContextService`


src/main/java/uk/gov/justice/digital/hmpps/oauth2server/security/UserContextApprovalHandler.kt
```diff
@@ -32,12 +36,9 @@ class UserContextApprovalHandler(private val userContextService: UserContextServ
   ): MutableMap<String, Any> {
 
     val userApprovalRequest = super.getUserApprovalRequest(authorizationRequest, userAuthentication)
-
-    if (isAzureAdUser(userAuthentication)) {
-      val userDetails = userAuthentication.principal as UserPersonDetails
-      val users = userContextService.discoverUsers(userDetails, authorizationRequest.scope)
-      userApprovalRequest["users"] = users
-    }
+    val userDetails = userAuthentication.principal as UserPersonDetails
```
tomsmyers, line 39: removing the `if` condition here fixes DT-1243. the effect of this change is that an empty list gets added to the userApprovalRequest, instead of no list at all. this causes the user selection screen to get skipped, instead of being shown with no users in the list.

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/service/UserContextService.kt
```diff
@@ -21,32 +21,27 @@ class UserContextService(
...
-      .filter { it.isNotEmpty() }
-      .flatten()
+      else -> emptyList()
+    }
   }
 
   private fun mapFromAzureAD(email: String, to: AuthSource): List<UserPersonDetails> = when (to) {
     delius -> deliusUserService.getDeliusUsersByEmail(email)
-    auth -> authUserService.findAuthUsersByEmail(email)
+    auth -> authUserService.findAuthUsersByEmail(email).filter { it.isVerified }
```
tomsmyers, line 44: this fixes DT-1242. the rest of the changes in this file are cosmetic.

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/security/UserContextApprovalHandler.kt
```diff
@@ -14,13 +14,17 @@ class UserContextApprovalHandler(private val userContextService: UserContextServ
     userAuthentication: Authentication,
   ): AuthorizationRequest {
 
-    val approvalRequest = super.checkForPreApproval(authorizationRequest, userAuthentication)
-    if (approvalRequest.isApproved && isAzureAdUser(userAuthentication)) {
-      // force all azure users down the approval route, the controller will work out what accounts are found etc.
-      approvalRequest.isApproved = false
+    // we have hijacked the UserContextApprovalHandler for our account selection process.
+    // we are purposefully not calling the super method, because if we deny the request
+    // based on unapproved scopes we do not currently have a way to explicitly approve it.
```
petergphillips, line 19: agreed, but that sounds correct to me?  If there are unapproved scopes then why should it be approved anyway?  Need to discuss I think...

petergphillips (approved):
LGTM

PR #610, opened 2020-10-21

petergphillips:


src/main/java/uk/gov/justice/digital/hmpps/oauth2server/nomis/model/AccountDetail.java
```diff
@@ -44,4 +44,19 @@ public AccountProfile getAccountProfile() {
         return AccountProfile.get(profile);
     }
 
+    public String getUsername() {
```
petergphillips, line 47: getters so can access from kotlin

src/test/java/uk/gov/justice/digital/hmpps/oauth2server/verify/VerifyMobileServiceIntTest.kt
```diff
@@ -18,13 +19,15 @@ import uk.gov.service.notify.NotificationClientApi
...
+  @Autowired
+  private lateinit var userTokenRepository: UserTokenRepository
   private val telemetryClient: TelemetryClient = mock()
   private val notificationClient: NotificationClientApi = mock()
   private lateinit var verifyMobileService: VerifyMobileService
 
   @BeforeEach
   fun setUp() {
-    verifyMobileService = VerifyMobileService(userRepository, null, telemetryClient, notificationClient, "templateId")
+    verifyMobileService = VerifyMobileService(userRepository, userTokenRepository, telemetryClient, notificationClient, "templateId")
```
petergphillips, line 30: had to change this test as don't want to pass `null` into the service any more

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/nomis/model/AccountDetail.java
tomsmyers: this is a behaviour change since previously these properties were private. what's the reason they need to be accessed from kotlin?

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/verify/ResetPasswordService.kt
```diff
@@ -0,0 +1,14 @@
+package uk.gov.justice.digital.hmpps.oauth2server.verify
+
+import uk.gov.justice.digital.hmpps.oauth2server.verify.ResetPasswordServiceImpl.NotificationClientRuntimeException
+import uk.gov.justice.digital.hmpps.oauth2server.verify.ResetPasswordServiceImpl.ResetPasswordException
+import java.util.Optional
+
+interface ResetPasswordService : PasswordService {
+  @Throws(NotificationClientRuntimeException::class)
+  fun requestResetPassword(usernameOrEmailAddress: String, url: String): Optional<String>
```
tomsmyers, line 9: there is extensive use of java optionals in the kotlin here. why not use kotlin's nullable types? i.e. `String?`

petergphillips: Completely agree.  I've tried to remove `stream()` and `Map.of` etc., but don't want to rewrite too much on a first pass.  `Optional` works okay with kotlin so intending to leave as is and then people can then improve the files gradually when they work on that area.

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/verify/ReferenceCodesService.kt
```diff
@@ -0,0 +1,50 @@
...
+  fun isValidEmailDomain(domain: String): Boolean {
+    if (StringUtils.isBlank(domain)) {
+      return false
+    }
+    val domainLower = domain.toLowerCase()
+    return try {
+      referenceCodesCache[ReferenceDomain.EMAIL_DOMAIN]
+        .map { it.replace("%".toRegex(), ".*").toLowerCase() }
+        .any { domainLower.matches(it.toRegex()) }
+    } catch (e: ExecutionException) {
```
andymarke, line 33: I'm guessing no need to catch now since it is all unchecked

andymarke (approved):
LGTM

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/verify/ResetPasswordService.kt
mikehalmamoj: When you get around to the above, I quite like the way @Mjwillis did this here: https://github.com/ministryofjustice/prisoner-offender-search/blob/main/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/services/IndexStatusService.kt#L91

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/nomis/model/AccountDetail.java
petergphillips: No behaviour change. The class has a `@Data` annotation on it which means that lombok did all the magic to expose getters and setters.  The kotlin pre-processor runs before the lombok one, so therefore have to delombok the getters in order for the code to be accessible from kotlin, until we convert the model class to kotlin.

PR #612, opened 2020-10-22

petergphillips:


src/main/java/uk/gov/justice/digital/hmpps/oauth2server/maintain/AuthUserGroupService.java
```diff
@@ -112,5 +110,13 @@ public AuthUserGroupException(final String field, final String errorCode) {
             this.field = field;
             this.errorCode = errorCode;
         }
+
+        public String getErrorCode() {
```
petergphillips, line 114: needed as kotlin pre processor occurs before lombok so doesn't see getters

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/resource/api/AuthUserController.kt
```diff
@@ -0,0 +1,527 @@
...
+  fun searchForUser(
+    @ApiParam(
+      value = "The username, email or name of the user.",
+      example = "j smith"
+    ) @RequestParam(required = false) name: String?,
+    @ApiParam(value = "The role of the user.") @RequestParam(required = false) role: String?,
+    @ApiParam(value = "The group of the user.") @RequestParam(required = false) group: String?,
+    @PageableDefault(sort = ["username"], direction = Sort.Direction.ASC) pageable: Pageable?,
+  ): ResponseEntity<Any> {
+    val users = authUserService.findAuthUsers(name, role, group, pageable!!).stream()
```
petergphillips, line 154: had to leave this in as it is operating on a `Page` object, not a `List`

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/resource/api/PublicKeyController.kt
```diff
@@ -0,0 +1,60 @@
...
+  fun getJwtPublicKey(): Map<String, Any> {
+    val formattedKey = getFormattedKey(publicKey)
+    return mapOf(
+      "formatted" to convertNewLinesToArray(formattedKey),
+      "encoded" to Base64.encodeBase64String(formattedKey.toByteArray())
+    )
+  }
+
+  private fun convertNewLinesToArray(formattedKey: String): Array<String> =
+    formattedKey.split("\n".toRegex()).dropLastWhile { it.isEmpty() }.toTypedArray()
```
petergphillips, line 35: need extra bit here - kotlin split adds extra `""` - see https://stackoverflow.com/questions/48697300/difference-between-kotlin-and-java-string-split-with-regex

electrified (approved):
LGTM

PR #615, opened 2020-10-23
DT-1248 force scopes and auto-approve scopes to always be the same

tomsmyers:
this is a pretty simple change, but is effective at blocking the "mismatched scopes" scenario which can cause un-auto-approved scopes to be granted to clients. 

i wanted to improve the error handling, by catching `InvalidScopeException` in the `/authorize` endpoint and redirecting back to the callback uri with error query params (a standard oauth2 pattern). but to implement correctly requires a lot of repetition of the spring authorize code, and comes out really messy. it would be easier if we could replace the framework's `AuthorizeEndpoint` bean with our own and subclass the built in implementation. but that is also fraught with difficulty. for now, this fixes the bug and is tidy from a client config perspective.


andymarke (approved):
LGTM

PR #619, opened 2020-10-26
DT-1224: converted more groovy tests to Kotlin

SimonMitchellMOJ:


src/test/java/uk/gov/justice/digital/hmpps/oauth2server/resource/api/UserControllerIntTest.kt
```diff
@@ -0,0 +1,236 @@
...
+import uk.gov.justice.digital.hmpps.oauth2server.resource.IntegrationTest
+
+@ExtendWith(DeliusExtension::class)
+class UserControllerIntTest : IntegrationTest() {
+
+  @Test
+  fun `User Me endpoint returns principal user data for client credentials grant`() {
+    webTestClient
+      .get().uri("/auth/api/user/me")
+      .headers(setAuthorisation("ITAG_USER"))
```
mikehalmamoj, line 16: I don't know if it's relevant to the test, but the original groovy test used client id `deliusnewtech` instead of `elite2apiclient`...

mikehalmamoj (approved):

src/test/java/uk/gov/justice/digital/hmpps/oauth2server/resource/api/UserControllerIntTest.kt
SimonMitchellMOJ: the client id doesn't matter for this endpoint as you are just retrieving details about you the user `ITAG_USER` so which ever clientid you are using it would still have the user in

PR #622, opened 2020-10-28
Dt 1252 java to kotlin

dnyanesh-sonwane:
More Java to kotlin conversion


src/main/java/uk/gov/justice/digital/hmpps/oauth2server/security/MaintainUserCheck.kt
```diff
@@ -0,0 +1,37 @@
...
+import uk.gov.justice.digital.hmpps.oauth2server.auth.model.User
+import uk.gov.justice.digital.hmpps.oauth2server.auth.repository.UserRepository
+
+@Service
+class MaintainUserCheck(
+  private val userRepository: UserRepository
+) {
+
+  companion object {
+    fun canMaintainAuthUsers(authorities: Collection<out GrantedAuthority>): Boolean {
```
mikehalmamoj, line 15: You don't need the `out` here - there should be a warning in Intellij

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/security/MaintainUserCheck.kt
```diff
@@ -0,0 +1,37 @@
...
+      return authorities.stream().map { obj: GrantedAuthority -> obj.authority }
+        .anyMatch { "ROLE_MAINTAIN_OAUTH_USERS".equals(it) }
```
mikehalmamoj, line 17: I know this is just a simple conversion and we're not going full idiomatic Kotlin (yet) - but in Kotlin you an use `==` which defaults to value equivalence rather than reference equivalence.

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/security/NomisH2AlterUserService.kt
```diff
@@ -0,0 +1,45 @@
...
+@Profile("!oracle")
+class NomisH2AlterUserService(
+  @Qualifier("dataSource") dataSource: DataSource,
+  passwordEncoder: PasswordEncoder,
+  staffUserAccountRepository: StaffUserAccountRepository,
+  userRepository: UserRepository
+) : NomisUserService(staffUserAccountRepository, userRepository) {
+
+  private val jdbcTemplate: JdbcTemplate = JdbcTemplate(dataSource)
+  private val encoder: PasswordEncoder = passwordEncoder
```
mikehalmamoj, line 23: This is redundant as it is part of the constructor - you can just change the constructor to have argument `private val encoder: PasswordEncoder`

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/maintain/AuthUserService.kt
```diff
@@ -175,7 +175,7 @@ class AuthUserService(
     val username = StringUtils.upperCase(usernameInput)
     val user = userRepository.findByUsernameAndMasterIsTrue(username)
       .orElseThrow { EntityNotFoundException("User not found with username $username") }
-    maintainUserCheck.ensureUserLoggedInUserRelationship(admin, authorities, user)
+    maintainUserCheck.ensureUserLoggedInUserRelationship(admin, authorities as Collection<out GrantedAuthority>, user)
```
mikehalmamoj, line 178: Rather than casting here, it might be better to change the signature of the method so that authorities are not nullable, e.g. `authorities: Collection<GrantedAuthority>`

mikehalmamoj (changes requested):
Mainly the redundant variable in NomisH2AlterUserService but the others are worth looking at too.

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/security/MaintainUserCheck.kt
dnyanesh-sonwane: Done

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/security/MaintainUserCheck.kt
dnyanesh-sonwane: Done

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/security/NomisH2AlterUserService.kt
dnyanesh-sonwane: Done

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/maintain/AuthUserService.kt
dnyanesh-sonwane: Done

mikehalmamoj (approved):

PR #630, opened 2020-11-03
Dt 1278 java to kotlin

dnyanesh-sonwane:
More java to kotlin


src/main/java/uk/gov/justice/digital/hmpps/oauth2server/config/OAuth2AuthorizationServerConfig.kt
```diff
@@ -0,0 +1,172 @@
...
+  fun tokenStore(): TokenStore? {
+    if (tokenStore == null) {
+      tokenStore = JwtTokenStore(accessTokenConverter())
+    }
+    return tokenStore
+  }
+
+  @Bean
+  @Primary
+  fun jdbcClientDetailsService(): JdbcClientDetailsService? {
```
mikehalmamoj, line 86: I don't think the return can be null?  Otherwise the `!!` would throw...

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/config/OAuth2AuthorizationServerConfig.kt
```diff
@@ -0,0 +1,172 @@
...
+
+  private fun userApprovalHandler(): UserApprovalHandler {
+    val approvalHandler = UserContextApprovalHandler(userContextService)
+    approvalHandler.setClientDetailsService(jdbcClientDetailsService())
+    approvalHandler.setRequestFactory(requestFactory())
+    approvalHandler.setTokenStore(tokenStore())
+    return approvalHandler
+  }
+
+  private fun requestFactory(): OAuth2RequestFactory? {
```
mikehalmamoj, line 133: Again not sure the return can be null?

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/config/SimpleSavedRequest.kt
```diff
@@ -0,0 +1,28 @@
...
+
+import org.springframework.security.web.savedrequest.SavedRequest
+import java.util.Locale
+import javax.servlet.http.Cookie
+
+/**
+ * Simple implementation of saved request that just uses the redirect url.
+ * Taken from https://github.com/AusDTO/spring-security-stateless/blob/master/src/main/java/au/gov/dto/springframework/security/web/savedrequest/SimpleSavedRequest.java
+ */
+class SimpleSavedRequest internal constructor(private val redirectUrl: String) : SavedRequest {
```
mikehalmamoj, line 11: Not relevant to this change but I noticed that there is now a `SimpleSavedRequest` in the `spring-security-web` jar, might want to switch to that one at some point?

mikehalmamoj (approved):

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/config/OAuth2AuthorizationServerConfig.kt
petergphillips: well intellij complains that it can be `null`.  In a multi threaded environment the field could be set to `null` after the assignment and before the `return`, hence why it moans.  In our case I didn't really care too much, so changed the return type to be nullable.  Could have equally done a
```
    return oAuth2RequestFactory!!
```
and not had a nullable return type.  Not sure it matters since in practise it will never be `null`...

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/config/OAuth2AuthorizationServerConfig.kt
petergphillips: see comment lower down, maybe it would be clearer to `!!` the return instead...

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/config/SimpleSavedRequest.kt
petergphillips: good plan, didn't realise that

src/main/java/uk/gov/justice/digital/hmpps/oauth2server/config/OAuth2AuthorizationServerConfig.kt
mikehalmamoj: Yeah seeing quite a few situations like this with Java interoperability.  Kotlin contracts sound like the answer but are still experimental... might be worth further investigation though.

PR #642, opened 2020-11-06
DT 1278 convert user class to kotlin

SimonMitchellMOJ:


src/main/java/uk/gov/justice/digital/hmpps/oauth2server/auth/model/User.kt
```diff
@@ -0,0 +1,301 @@
...
+  var mfaPreference = MfaPreferenceType.EMAIL
+
+  @ElementCollection
+  @CollectionTable(name = "USER_CONTACT", joinColumns = [JoinColumn(name = "user_id")])
+  val contacts: MutableSet<Contact> = mutableSetOf()
+
+  @OneToMany(mappedBy = "user", cascade = [CascadeType.ALL], orphanRemoval = true)
+  val tokens: MutableSet<UserToken> = mutableSetOf()
+
+  private constructor (
```
petergphillips, line 110: constructor not needed, will remove in subsequent PR

JamieDraperUK (approved):

PR #646, opened 2020-11-06
DT-1278: moved all files (and test files) from java folders to kotlin…

SimonMitchellMOJ:
… folders as no java files left to convert


petergphillips (approved):
LGTM

PR #647, opened 2020-11-09
🐛 fixed an issue of empty auth roles resulting in incorrect service a…

dnyanesh-sonwane:
DT-1293
- Fixed issue of empty authorisation roles causing services not being displayed correctly

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/resource/ServicesController.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/resource/ServicesController.kt
@@ -46,9 +46,11 @@ class ServicesController(
     val userDetails = authentication.principal as UserPersonDetails
     val telemetryMap = mapOf("username" to userDetails.username, "code" to service.code)
     if (newService) {
+      replaceBlankAuthorisedRolesWithNull(service)
       authServicesService.addService(service)
       telemetryClient.trackEvent("AuthServiceDetailsAdd", telemetryMap, null)
     } else {
+      replaceBlankAuthorisedRolesWithNull(service)
       authServicesService.updateService(service)
       telemetryClient.trackEvent("AuthServiceDetailsUpdate", telemetryMap, null)
     }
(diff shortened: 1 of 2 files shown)


mikehalmamoj (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/resource/ServicesController.kt
```diff
@@ -46,9 +46,11 @@ class ServicesController(
     val userDetails = authentication.principal as UserPersonDetails
     val telemetryMap = mapOf("username" to userDetails.username, "code" to service.code)
     if (newService) {
+      replaceBlankAuthorisedRolesWithNull(service)
```
petergphillips, line 49: duplicated line - should really have gone outside the `if else` block

src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/resource/ServicesController.kt
```diff
@@ -64,4 +66,8 @@ class ServicesController(
     telemetryClient.trackEvent("AuthServiceDetailsDeleted", telemetryMap, null)
     return "redirect:/ui/services"
   }
+
+  private fun replaceBlankAuthorisedRolesWithNull(service: Service) {
+    service.authorisedRoles = if (service.authorisedRoles.isNullOrBlank()) null else service.authorisedRoles
```
petergphillips, line 71: this is a bit naff - reassigning an field on `service` like this.  Doesn't cope with a newline in the textarea either.

src/test/kotlin/uk/gov/justice/digital/hmpps/oauth2server/resource/ServicesControllerTest.kt
```diff
@@ -87,6 +87,19 @@ class ServicesControllerTest {
         null
       )
     }
+    @Test
+    fun `edit service - edit service blank authorised roles`() {
+      val service = Service(code = "editcode", name = "", description = "", url = "", authorisedRoles = "")
+      val url = controller.editService(authentication, service)
+      assertThat(url).isEqualTo("redirect:/ui/services")
+      assertThat(service.authorisedRoles).isNull()
```
petergphillips, line 95: this line of the test relies on the input `service` being mutated, really should check the parameter to the service instead as they could be different.

PR #649, opened 2020-11-10
Dt 1293 empty auth roles fix

dnyanesh-sonwane:
Dt 1293 fix for when authorisation roles are sent in as new lines or empty string


petergphillips (approved):
LGTM

PR #652, opened 2020-11-11
Dt 1247 use ojdbcjar gradle dependency

dnyanesh-sonwane:
deleted lib folder ( including ojdbc jar) and added it as a gradle dependency


petergphillips (approved):
LGTM

PR #653, opened 2020-11-11
DT-1305: 🐛 Add missing child groups foreign key

petergphillips:


JamieDraperUK (approved):
LGTM - Is it worth going through the trouble trying to write a test for the bug?

PR #656, opened 2020-11-13
Dt 1285 group amendments

dnyanesh-sonwane:
amendments to the groups


src/main/resources/db/auth/V5_06__group_add_amend_remove.sql
```diff
@@ -0,0 +1,22 @@
+update groups  set group_name='PECS Court BARROW IN FURNESS COMBINED COURT'
```
petergphillips, line 1: shouldn't all be in capitals

petergphillips (approved):
LGTM

src/main/resources/db/auth/V5_06__group_add_amend_remove.sql
dnyanesh-sonwane: changed to camel case as PECS Court Barrow In Furness Combined Court , so it is in line with other group names

PR #669, opened 2020-11-20
DT-1054: 🐛 Update auth source to reflect last login

petergphillips:


src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/auth/model/User.kt
```diff
@@ -46,11 +46,11 @@ class User(
   private var enabled: Boolean = false,
 
   /**
-   * Whether we are masters of this user record here in auth
+   * Source of last login information
    */
   @Column(name = "source", nullable = false)
   @Enumerated(EnumType.STRING)
-  val source: AuthSource,
+  var source: AuthSource,
```
electrified, line 53: I am thinking that AuthSource should remain an immutable field? 
If the auth source is different, it should be an additional row in the table?

electrified (approved):

PR #673, opened 2020-11-23
DT-1297 - Update AppInsights agent configuration

electrified:
Stop app insights from attempting to write to a config file in-container

I've been unable to disable it from logging JDBC calls, so have instead used the log processor to redact the ALTER USER statements which would contain the users password on Oracle


petergphillips (approved):
LGTM

AI-Agent.xml
```diff
@@ -1,11 +0,0 @@
-<?xml version="1.0" encoding="utf-8"?>
-<ApplicationInsightsAgent>
-  <Instrumentation>
-    <BuiltIn enabled="true">
-      <HTTP enabled="true" W3C="true"/>
```
petergphillips, line 5: worth checking that we still get w3c tracing, had to explicitly enable it here

electrified: The docs say w3c tracing doesn't require any config now, but I will check it. https://docs.microsoft.com/en-us/azure/azure-monitor/app/correlation#java-30-agent

PR #676, opened 2020-11-24
DT-1297 - Add app version and role name to app insights tracing

electrified:
The app version is now available as `customDimensions.application_Version`. vs `application_Version`. Not 100% happy with this but have not found a way to do this yet.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,6 +14,8 @@ RUN apt-get update && \
     apt-get install -y curl && \
     rm -rf /var/lib/apt/lists/*
 
+ARG BUILD_NUMBER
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -14,6 +14,8 @@ RUN apt-get update && \
     apt-get install -y curl && \
     rm -rf /var/lib/apt/lists/*
 
+ARG BUILD_NUMBER
```
electrified, line 17: ARGs get reset after every FROM so needs to be redefined

petergphillips (approved):
LGTM

PR #679, opened 2020-11-26
DT-1217 Improve manage account details for Azure AD users

electrified:


src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/security/UserRetriesService.kt
```diff
@@ -36,6 +37,12 @@ class UserRetriesService(
       user.email = userPersonDetails.email
       user.verified = true
     }
+
+    if (userPersonDetails is AzureUserPersonDetails) {
```
petergphillips, line 41: might be nicer to add this to the existing `if` rather than as a separate `if` block

src/main/resources/templates/account/accountDetails.html
```diff
@@ -11,7 +11,7 @@
   <div role="alert">
     <h1 class="govuk-heading-l">Your account details</h1>
   </div>
-  <div>
+  <div th:if="${authUser.authSource != 'azuread'}" >
```
petergphillips, line 14: might be better to do a `th:unless` rather than a negated `th:if`

src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/resource/account/AccountController.kt
```diff
@@ -5,17 +5,29 @@ import org.springframework.stereotype.Controller
...
-class AccountController(private val userService: UserService) {
+class AccountController(private val userService: UserService, private val userContextService: UserContextService) {
   @GetMapping("/account-details")
   fun accountDetails(authentication: Authentication): ModelAndView {
     val username = authentication.name
     val user = userService.findMasterUserPersonDetails(username).orElseThrow()
     val authUser = userService.getUserWithContacts(username)
+    val linkedAccounts = userContextService.discoverUsers(user, emptySet())
+      .map {
+        LinkedAccountModel(it.authSource.toUpperCase(), it.username)
```
petergphillips, line 19: `authSource` being a `String` on `UserPersonDetails`, means that we can't use the enum directly.  We'd like to start calling the `auth` users `External` as well, might be good to have a description on the enum and map it to `DPS`, `Delius` and `External`?

petergphillips: obviously subsequent PR

src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/resource/account/AccountController.kt
```diff
@@ -5,17 +5,29 @@ import org.springframework.stereotype.Controller
...
 
 @Controller
-class AccountController(private val userService: UserService) {
+class AccountController(private val userService: UserService, private val userContextService: UserContextService) {
   @GetMapping("/account-details")
   fun accountDetails(authentication: Authentication): ModelAndView {
     val username = authentication.name
     val user = userService.findMasterUserPersonDetails(username).orElseThrow()
     val authUser = userService.getUserWithContacts(username)
+    val linkedAccounts = userContextService.discoverUsers(user, emptySet())
```
petergphillips, line 17: now that everything is kotlin, we could start doing nice things like having default values on the methods e.g. second parameter of `discoverUsers`

petergphillips (approved):
LGTM

src/main/resources/templates/account/accountDetails.html
```diff
@@ -33,7 +33,7 @@ <h1 class="govuk-heading-l">Your account details</h1>
         </a>
       </dd>
     </div>
-    <div class="govuk-summary-list__row">
+    <div class="govuk-summary-list__row" th:if="${authUser.authSource != 'azuread'}">
```
petergphillips, line 36: missed out the `unless` :grinning:

src/main/resources/templates/account/accountDetails.html
```diff
@@ -65,7 +65,7 @@ <h1 class="govuk-heading-l">Your account details</h1>
       </dd>
       <dd class="govuk-summary-list__actions">
         <a class="govuk-link govuk-link--no-visited-state" th:href="@{/existing-email}"
-           th:if="${authUser.authSource != 'delius'}"
+           th:if="${authUser.authSource != 'delius' && authUser.authSource != 'azuread'}"
            data-qa="changeEmail">
           <th:block th:text="${#strings.isEmpty(authUser.email)} ? 'Add' : 'Change'">
             Change
```
petergphillips, line 71: and here :grin:

petergphillips (approved):
LGTM

PR #694, opened 2020-12-08
PAT-1454: add SOC HQ user role

connormaglynn:


src/main/resources/db/auth/V5_12__add_soc_hq_role.sql
```diff
@@ -0,0 +1,2 @@
+INSERT INTO roles (role_id, role_code, role_name)
+VALUES (newid(), 'SOC_HQ', 'SOC HQ User');
```
petergphillips, line 2: Would be good to expand out `HQ` in the role name - what does `HQ` stand for?

src/main/resources/db/auth/V5_12__add_soc_hq_role.sql
```diff
@@ -0,0 +1,2 @@
+INSERT INTO roles (role_id, role_code, role_name)
```
petergphillips, line 1: who should get this role - normally should go hand in hand with insert into `GROUP_ASSIGNABLE_ROLE` so that, even if they don't get assigned the role by default, it can be assigned by a group manager (otherwise only auth admins can assign the role)

src/main/resources/db/auth/V5_12__add_soc_hq_role.sql
```diff
@@ -0,0 +1,2 @@
...
+VALUES (newid(), 'SOC_HQ', 'SOC Head Quarters User');
```
petergphillips, line 2: sorry to be picky - shouldn't this be `Headquarters` instead - what is the main hub of SOC called?

petergphillips (approved):
LGTM

PR #697, opened 2020-12-11
DT-1405: 🔊 Silence curl from showing progress

petergphillips:


user-load.sh
```diff
@@ -86,7 +86,7 @@ while IFS=, read -r -a row; do
   printf -v groups '\"%s\",' "${row[@]:4}"
 
   # Create the user
-  if ! output=$(curl -X PUT "$HOST/auth/api/authuser/$user?enforceUniqueEmail=true" -H "$AUTH_TOKEN_HEADER" -H "Content-Type: application/json" \
+  if ! output=$(curl -sS -X PUT "$HOST/auth/api/authuser/$user?enforceUniqueEmail=true" -H "$AUTH_TOKEN_HEADER" -H "Content-Type: application/json" \
```
petergphillips, line 89: `-sS` is `curl --silent --show-error` so will be silent, but still show errors if there are any

SimonMitchellMOJ (approved):
LGTM

PR #700, opened 2020-12-16
DT-1429: ✨ Change email address also changes username

petergphillips:


src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/resource/VerifyEmailController.kt
```diff
@@ -111,7 +111,6 @@ class VerifyEmailController(
         modelAndView.addObject("verifyLink", verifyLink)
       }
       modelAndView.addObject("email", chosenEmail)
-      modelAndView
```
petergphillips, line 114: unnecessary since `addObject` already returns it

mikehalmamoj (approved):

PR #702, opened 2020-12-17
DT-1429: 🐛 Change jwt after username amendment

petergphillips:


src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/verify/VerifyEmailService.kt
```diff
@@ -112,7 +112,7 @@ class VerifyEmailService(
       throw e
     }
     userRepository.save(user)
-    return verifyLink
+    return LinkAndEmail(verifyLink, email!!)
```
petergphillips, line 115: can `!!` as the email has been verified in the validate method and exception thrown there

src/test/kotlin/uk/gov/justice/digital/hmpps/oauth2server/integration/ChangeEmailSpecification.kt
```diff
@@ -135,6 +135,9 @@ class ChangeEmailSpecification : AbstractDeliusAuthSpecification() {
       .inputAndConfirmNewEmail("auth_change_email_new@justice.gov.uk")
     verifyEmailSentPage
       .isAt()
+    // check account details page displays correctly
+    goTo(accountDetailsPage).isAtPage()
```
petergphillips, line 139: this was previously blowing up

JamieDraperUK (approved):
LGTM

PR #703, opened 2020-12-17
DT-1429: 🐛 Fix changing email for unverified accounts

petergphillips:


src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/maintain/AuthUserService.kt
```diff
@@ -191,11 +191,11 @@ class AuthUserService(
     val user = userRepository.findByUsernameAndMasterIsTrue(username)
       .orElseThrow { EntityNotFoundException("User not found with username $username") }
     maintainUserCheck.ensureUserLoggedInUserRelationship(admin, authorities, user)
-    if (user.verified) {
+    if (user.password != null) {
```
petergphillips, line 194: we want initial users here, but was catching users that had changed their email without verifying it

JamieDraperUK (approved):
LGTM

PR #704, opened 2020-12-17
Dt 1427 create user by email

rgforsyth:


src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/maintain/AuthUserService.kt
```diff
@@ -257,10 +292,20 @@ class AuthUserService(
...
+
     validate(firstName, lastName)
     verifyEmailService.validateEmailAddress(email, emailType)
   }
 
+  @Throws(CreateUserException::class, VerifyEmailException::class)
+  private fun validate(email: String?, firstName: String?, lastName: String?, emailType: EmailType) {
+    validate(firstName, lastName)
+
+    if (email.isNullOrBlank() || email.length > MAX_LENGTH_EMAIL) throw CreateUserException("username", "maxlength")
```
petergphillips, line 304: would be better to do this validation in `verifyEmailService.validateEmailAddress`, surprised it isn't there already

src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/resource/api/AuthUserController.kt
```diff
@@ -186,6 +187,90 @@ class AuthUserController(
     return roles.map { AuthUserRole(it) }
   }
 
+  @PostMapping("/api/authuser/")
```
petergphillips, line 190: would be better to have a `/api/authuser/create` endpoint instead, I wouldn't expect `POST`ing here to create a new user

src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/resource/api/AuthUserController.kt
```diff
@@ -186,6 +187,90 @@ class AuthUserController(
...
+  fun createUserByEmail(
+    @ApiParam(value = "Details of the user to be created.", required = true) @RequestBody createUser: CreateUser,
+    @ApiIgnore request: HttpServletRequest,
+    @ApiIgnore authentication: Authentication,
+  ): ResponseEntity<Any> {
+
+    val email = EmailHelper.format(createUser.email)
+
+    val user =
+      if (StringUtils.isNotBlank(email)) userService.findMasterUserPersonDetails(StringUtils.trim(email)) else Optional.empty<Any>()
```
petergphillips, line 217: Not your fault since original code was just converted as is, but not very kotlin like.  Would be better to do something like
```
    val user = email?.let { userService.findMasterUserPersonDetails(email).orElse(null) }
```
and then check `user` isn't null

src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/resource/api/AuthUserController.kt
```diff
@@ -186,6 +187,90 @@ class AuthUserController(
...
+    } catch (e: CreateUserException) {
+      log.info(
+        "Create user failed for user {} for field {} with reason {}",
+        createUser.email,
+        e.field,
+        e.errorCode
+      )
+      ResponseEntity.badRequest().body(
+        ErrorDetail(
+          String.format("%s.%s", e.field, e.errorCode),
```
petergphillips, line 262: better to use string interpolation here

src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/resource/api/AuthUserController.kt
```diff
@@ -186,6 +187,90 @@ class AuthUserController(
...
+          String.format("%s failed validation", e.field),
+          e.field
+        )
+      )
+    } catch (e: VerifyEmailException) {
+      log.info("Create user failed for user {} for field email with reason {}", email, e.reason)
+      ResponseEntity.badRequest()
+        .body(ErrorDetail(String.format("email.%s", e.reason), "Email address failed validation", "email"))
```
petergphillips, line 270: string interpolation preferred

src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/resource/api/AuthUserController.kt
```diff
@@ -186,6 +187,90 @@ class AuthUserController(
...
+  ): ResponseEntity<Any> {
+
+    val email = EmailHelper.format(createUser.email)
+
+    val user =
+      if (StringUtils.isNotBlank(email)) userService.findMasterUserPersonDetails(StringUtils.trim(email)) else Optional.empty<Any>()
+
+    // check that we're not asked to create a user that is already in nomis, auth or delius
+    if (user.isPresent) {
+      return ResponseEntity.status(HttpStatus.CONFLICT)
```
petergphillips, line 221: don't change it, but this is the old way of writing services.  We are now trying to return a type and exceptions thrown and handled by the global exception handler instead.  Makes lots of this code simpler as there is a standard place to handle and means that the user checks can be moved down to the service instead

src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/resource/api/AuthUserController.kt
petergphillips: I guess the other option would be to keep it as a `PUT` with the email as a parameter

petergphillips (dismissed):
Added some comments, happy to be looked at in subsequent PR...

petergphillips (approved):
LGTM

PR #706, opened 2020-12-18
Dt 1427 review comments

rgforsyth:


src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/resource/api/AuthUserController.kt
```diff
@@ -259,15 +257,15 @@ class AuthUserController(
...
-          String.format("%s.%s", e.field, e.errorCode),
-          String.format("%s failed validation", e.field),
+          "${e.field}.${e.errorCode}",
+          "${e.field} failed validation",
           e.field
         )
       )
     } catch (e: VerifyEmailException) {
-      log.info("Create user failed for user {} for field email with reason {}", email, e.reason)
+      log.info("Create user failed for user {} for field email with reason $email")
```
petergphillips, line 266: should be
```
      log.info("Create user failed for user $email for field email with reason ${e.reason}")
```

src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/resource/api/AuthUserController.kt
```diff
@@ -259,15 +257,15 @@ class AuthUserController(
...
       ResponseEntity.badRequest()
-        .body(ErrorDetail(String.format("email.%s", e.reason), "Email address failed validation", "email"))
+        .body(ErrorDetail(String.format("email.${e.reason}", e.reason), "Email address failed validation", "email"))
```
petergphillips, line 268: you've left the `String.format` in here

src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/verify/VerifyEmailService.kt
```diff
@@ -173,6 +173,7 @@ class VerifyEmailService(
     if (email == null) {
       throw VerifyEmailException("blank")
     }
+    if (email.isNullOrBlank() || email.length > MAX_LENGTH_EMAIL) throw VerifyEmailException("maxlength")
```
petergphillips, line 176: would be better to change the
```
    if (email == null) {
```
check above to be
```
if (email.isNullOrBlank()
```
and then this one can just be about the length, otherwise they will receive a maxlength message if it is blank

src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/resource/api/AuthUserController.kt
```diff
@@ -263,9 +263,9 @@ class AuthUserController(
         )
       )
     } catch (e: VerifyEmailException) {
-      log.info("Create user failed for user {} for field email with reason $email")
+      log.info("Create user failed for user {} for field email with reason ${e.reason}")
```
petergphillips, line 266: oops my bad - `log` messages shouldn't use interpolation since that is always evaluated.  Could you please revert this back to what it was doing before

petergphillips (approved):
LGTM

PR #709, opened 2020-12-21
DT-1447: ✨ Add in current mfa status to session cookie

petergphillips:


src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/security/JwtAuthenticationSuccessHandler.kt
```diff
@@ -75,18 +76,15 @@ open class JwtAuthenticationSuccessHandler(
   ) {
     val optionalAuth = jwtCookieHelper.readValueFromCookie(request)
       .flatMap { jwtAuthenticationHelper.readUserDetailsFromJwt(it) }
-    val jwt =
-      optionalAuth.map { udi: UserDetailsImpl -> jwtAuthenticationHelper.createJwtWithId(authentication, udi.jwtId) }
-        .orElseGet { jwtAuthenticationHelper.createJwt(authentication) }
+    val passedMfa = optionalAuth.map { it.passedMfa }.orElse(false)
```
petergphillips, line 79: need to copy across any previous mfa status to new cookie, this method is used when the user has changed their details in the session (e.g. name) and we need to generate a new cookie for them with the updated details

andrewrlee (approved):
LGTM

PR #715, opened 2021-01-06
DT-1470: ⬆️🚧 Try out application insights agent v3 (beta)

petergphillips:


helm_deploy/hmpps-auth/templates/_envs.tpl
```diff
@@ -19,6 +19,8 @@ env:
       secretKeyRef:
         key: APPLICATION_INSIGHTS_IKEY
         name: {{ template "app.name" . }}
+  - name: APPLICATIONINSIGHTS_CONNECTION_STRING
+    value: "InstrumentationKey=$(APPLICATION_INSIGHTS_IKEY)"
```
petergphillips, line 23: This way we can keep the secret just containing the ikey in AWS and then create the connection string variable from it

src/test/kotlin/uk/gov/justice/digital/hmpps/oauth2server/utils/JwtAuthHelper.kt
```diff
@@ -23,16 +23,30 @@ class JwtAuthHelper(
 ) {
   private val keyPair: KeyPair
 
-  fun createJwt(parameters: JwtParameters): String {
-    val claims = mutableMapOf<String, Any>("user_name" to parameters.username, "client_id" to "elite2apiclient")
-    parameters.roles?.let { claims["authorities"] = parameters.roles }
-    parameters.scope?.let { claims["scope"] = parameters.scope }
-    parameters.additionalClaims?.let { claims.putAll(parameters.additionalClaims) }
+  fun createJwt(parameters: JwtParameters): String = createJwt(
```
petergphillips, line 26: don't really need a data class to pass through the parameters (hang up when was in java), but left here as would have caused lots of file changes

applicationinsights.json
```diff
@@ -0,0 +1,46 @@
...
+    "service.version": "${BUILD_NUMBER}"
+  },
+  "preview": {
+    "processors": [
+      {
+        "type": "attribute",
+        "include": {
+          "matchType": "regexp",
+          "spanNames": [
+            "^(.*?)ALTER USER(.*?)$"
```
petergphillips, line 15: will start adding sql to `dependencies`, but don't want passwords exposed

JamieDraperUK (approved):
lgtm

PR #721, opened 2021-01-07
Tidying up the upload script

andrewrlee:
Renaming misleading variable
Correcting upload script instructions
Removing adding of vary role as this will never be used again


petergphillips (approved):
LGTM

PR #726, opened 2021-01-11
add groups if username already present and can be determined

rgforsyth:


user-load.sh
```diff
@@ -69,6 +69,17 @@ TOKEN=$(echo "$TOKEN_RESPONSE" | jq -er .access_token)
 
 AUTH_TOKEN_HEADER="Authorization: Bearer $TOKEN"
 
+addGroup() {
+  local user=$1
+  local group=$2
+  if [[ "$group" != "" && ! "$group" =~ ^[,]*$ ]]; then
+    curl -s -X PUT "$HOST/auth/api/authuser/$user/groups/$group" -H "Content-Length: 0" -H "$AUTH_TOKEN_HEADER"
```
petergphillips, line 76: could do with a `S` option here too like the others have now got
>        -S, --show-error
>              When used with -s, --silent, it makes curl show an error message if it fails.

user-load.sh
```diff
@@ -69,6 +69,17 @@ TOKEN=$(echo "$TOKEN_RESPONSE" | jq -er .access_token)
...
+    if [[ $? -ne 0 ]]; then
```
petergphillips, line 77: worth installing shellcheck locally, think it will warn on this line and come up with a suggestion to combine the `curl` into it

petergphillips (dismissed):
LGTM

mikehalmamoj (approved):

PR #727, opened 2021-01-11
add ROLE_OASYS_READ_ONLY to local interventions client to allow use of offender-assessments-api

tomsmyers:
small change to enable local dev using offender-assessments-api with the interventions client credentials.


petergphillips (approved):
LGTM

PR #733, opened 2021-01-15
DT1505: fix back link to return to the correct page

rgforsyth:


petergphillips (approved):
LGTM

petergphillips (changes requested):
think you need `<script>` wrapping in `th:if="${showBack}"` too

PR #736, opened 2021-01-19
DT-1512 username to be in lowercase(in email text)  when email is use…

dnyanesh-sonwane:
…rname

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/verify/ResetPasswordServiceImpl.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/verify/ResetPasswordServiceImpl.kt
@@ -182,7 +182,7 @@ class ResetPasswordServiceImpl(
 
   private fun sendPasswordResetEmail(user: User) {
     // then the reset token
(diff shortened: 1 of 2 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/verify/ResetPasswordServiceImpl.kt
```diff
@@ -182,7 +182,7 @@ class ResetPasswordServiceImpl(
 
   private fun sendPasswordResetEmail(user: User) {
     // then the reset token
-    val username = user.username
+    val username = if (StringUtils.contains(user.username, "@")) user.username.toLowerCase() else user.username
```
petergphillips, line 185: kotlin does have a `contains` on `String` so don't need to call `StringUtils` for it

petergphillips (approved):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/verify/ResetPasswordServiceImpl.kt
dnyanesh-sonwane: Done

petergphillips (approved):
LGTM

PR #741, opened 2021-01-25
DT-1505: updated backlink to verify the domain is held in the client details

SimonMitchellMOJ:


src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/resource/account/BackLinkHandler.kt
```diff
@@ -0,0 +1,74 @@
...
+  }
+
+  fun validateRedirect(
+    client: String,
+    redirect: String,
+  ): Boolean {
+
+    val uri = URI(redirect)
+    val url = when {
+      uri.port == 443 -> "${uri.scheme}://${uri.host}"
```
petergphillips, line 41: 443 and -1 are the same result, so could do an `if` block instead.  Wondering about 80 too, depending on whether the ssl is terminated at nginx?

src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/resource/account/BackLinkHandler.kt
```diff
@@ -0,0 +1,74 @@
...
+      uri.port != -1 -> "${uri.scheme}://${uri.host}:${uri.port}"
+      else -> "${uri.scheme}://${uri.host}"
+    }
+
+    // If we have asked for a redirect, check it is valid for the client
+    val clientDetails = clientDetailsService.loadClientByClientId(client)
+
+    if (clientDetails != null && !CollectionUtils.isEmpty(clientDetails.registeredRedirectUri)) {
```
petergphillips, line 49: do you have to use `CollectionUtils` here, doesn't kotlin provide an `isEmpty`?

src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/resource/account/BackLinkHandler.kt
```diff
@@ -0,0 +1,74 @@
...
+  }
+
+  private fun responseRedirectedOnValidRedirect(
+    redirect: String,
+    clientDetails: ClientDetails
+  ): Boolean {
+    try {
+
+      redirectResolver.resolveRedirect(redirect, clientDetails)
+      return true
```
petergphillips, line 68: looking slightly messy here, would be better to `return try` instead

petergphillips (approved):
LGTM

PR #752, opened 2021-02-02
DT-1559: add endpoint to delete group

SimonMitchellMOJ:


src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/auth/model/Group.kt
```diff
@@ -29,6 +29,9 @@ class Group(
   @OneToMany(mappedBy = "group", cascade = [CascadeType.ALL], orphanRemoval = true)
   val children: MutableSet<ChildGroup> = mutableSetOf()
 
+  // @ManyToMany(mappedBy = "groups", cascade = [CascadeType.DETACH])
```
petergphillips, line 32: unnecessary commented out code?

src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/maintain/GroupsService.kt
```diff
@@ -40,31 +44,67 @@ class GroupsService(
...
+  @Transactional(transactionManager = "authTransactionManager")
+  @Throws(GroupNotFoundException::class, GroupHasChildGroupException::class)
+  fun deleteGroup(username: String, groupCode: String) {
+    val group = groupRepository.findByGroupCode(groupCode) ?: throw
+    GroupNotFoundException("delete", groupCode, "notfound")
+
+    when {
+      group.children.isEmpty() -> {
+        removeUsersFromGroup(groupCode, username)
+        groupRepository.deleteByGroupCode(groupCode)
```
petergphillips, line 75: this is slightly odd.  You've got the `group` already so should just call `delete` or `remove` on that rather than finding it again.

petergphillips (approved):
LGTM

PR #753, opened 2021-02-03
Faster JVM startup.

mattops:
Namespace LimitRange set to much higher CPU limit of 2000ms - java now starts much faster so tightening up the liveness/readiness probes.


andrewrlee (approved):
LGTM

PR #756, opened 2021-02-09
DT-1602: 2fa on add/change backup email

SimonMitchellMOJ:


src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/resource/VerifyEmailController.kt
```diff
@@ -90,6 +90,7 @@ class VerifyEmailController(
     @RequestParam(required = false) candidate: String,
     @RequestParam email: String?,
     @RequestParam emailType: EmailType,
+    @RequestParam token: String?,
```
petergphillips, line 93: this is a bit weird - the `token` isn't used in the post mapping.  Doesn't that mean that an attacker could bypass the 2FA challenge by `POST`ing the details to this endpoint?

src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/resource/AccountMfaController.kt
```diff
@@ -0,0 +1,129 @@
...
+
+  @GetMapping("/account/mfa-challenge")
+  fun serviceMfaChallengeRequest(
+    authentication: Authentication,
+    @RequestParam contactType: String,
+  ): ModelAndView {
+    // issue token to current mfa preference
+    val mfaData = mfaService.createTokenAndSendMfaCode(authentication.name)
+    val codeDestination = mfaService.getCodeDestination(mfaData.token, mfaData.mfaType)
+    val modelAndView = ModelAndView("mfaChallengeAD", "token", mfaData.token)
```
petergphillips, line 40: would prefer a longer name here to be clearer e.g. `accountDetailsMfaChallenge` or `mfaChallengeAccountDetails`

petergphillips (approved):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/resource/VerifyEmailController.kt
SimonMitchellMOJ: will check that out

PR #758, opened 2021-02-10
DT-1602: Added token check for 2fa on secondary email

SimonMitchellMOJ:


src/main/resources/messages.properties
```diff
@@ -99,4 +101,5 @@ userdetails.lastName.maxlength=Your last name must be 50 characters or less
 userdetails.lastName.invalid=Your last name cannot contain < or > characters
 existingpassword.invalid=Your password is incorrect. You will be locked out if you enter the wrong details {0} times.
 existingpassword.required=Enter your current password
-existingpassword.deliusdown=Delius is experiencing issues. Please try later if you are attempting to sign in using your Delius credentials.
+existingpassword.deliusdown=Delius is experiencing issues. Please try later if you are attempting to sign in using your Delius credentials
```
petergphillips, line 104: removed full stop at end here?  Normally if we have more than one sentence both end with full stops

src/main/resources/messages.properties
```diff
@@ -39,6 +39,8 @@ verifyemail.gsi=All gsi.gov.uk have now been migrated to a justice.gov.uk domain
 verifyemail.other=An error occurred during sending the verification email. Please try again.
 verifyemail.nosecondaryemail=No backup email found
 verifyemail.duplicate=There is already an account with this email address.  Please sign in with that email address instead.
+verifyemail.token.invalid=Security token is invalid, Please try updating your email again.
+verifyemail.token.expired=Security token has expired, Please try updating your email again.
```
petergphillips, line 43: don't need capital `P` in Please and don't need full stop.  Worth adding a `The` to start of both again.  I think the link on the account details is `Change` rather than `Update` so should be `changing your email again`?

petergphillips (approved):
LGTM

PR #763, opened 2021-02-12
DT-1603: add 2fa on add/change mobile

SimonMitchellMOJ:


src/main/resources/messages.properties
```diff
@@ -39,8 +39,8 @@ verifyemail.gsi=All gsi.gov.uk have now been migrated to a justice.gov.uk domain
 verifyemail.other=An error occurred during sending the verification email. Please try again.
 verifyemail.nosecondaryemail=No backup email found
 verifyemail.duplicate=There is already an account with this email address.  Please sign in with that email address instead.
-verifyemail.token.invalid=The security token is invalid. Please try changing your email again.
-verifyemail.token.expired=The security token has expired. Please try changing your email again.
+verifyemail.token.invalid=The security token is invalid. Please try changing your backup email again.
```
petergphillips, line 42: is this only triggered for backup emails?

src/main/resources/templates/account/accountDetails.html
```diff
@@ -114,7 +113,7 @@ <h1 class="govuk-heading-l">Your account details</h1>
       <dt class="govuk-summary-list__key">
         Backup email
       </dt>
-      <dd class="govuk-summary-list__value" data-qa="secondaryEmail" th:text="${authUser.secondaryEmail}">
+      <dd class="govuk-summary-list__value" id="secondaryEmail" data-qa="secondaryEmail" th:text="${authUser.secondaryEmail}">
```
petergphillips, line 116: shouldn't really need `id`s here if we have a `data-qa` attribute?

petergphillips (approved):
LGTM

src/main/resources/messages.properties
SimonMitchellMOJ: At the moment yes, It will be updated when I do the primary email

PR #771, opened 2021-02-22
DT-1603: updated account details so that 2fa code is not resent to user…

SimonMitchellMOJ:
…every time page reload/redirect with error


src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/resource/MfaControllerAccountDetails.kt
```diff
@@ -30,24 +30,36 @@ class MfaControllerAccountDetails(
...
+      // issue token to current mfa preference
+      val mfaData = mfaService.createTokenAndSendMfaCode(authentication.name)
+      val codeDestination = mfaService.getCodeDestination(mfaData.token, mfaData.mfaType)
+      val modelAndView = ModelAndView("mfaChallengeAccountDetails", "token", mfaData.token)
+        .addObject("mfaPreference", mfaData.mfaType)
+        .addObject("codeDestination", codeDestination)
+        .addObject("contactType", contactType)
+      if (smokeTestEnabled) modelAndView.addObject("smokeCode", mfaData.code)
+      return modelAndView
+    } else {
```
petergphillips, line 50: don't need an `else` after `return` from the `if`

src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/resource/MfaControllerAccountDetails.kt
```diff
@@ -30,24 +30,36 @@ class MfaControllerAccountDetails(
...
+      val codeDestination = mfaService.getCodeDestination(token!!, mfaPreference!!)
+      val modelAndView = ModelAndView("mfaChallengeAccountDetails", "token", token)
+        .addObject("mfaPreference", mfaPreference)
+        .addObject("codeDestination", codeDestination)
+        .addObject("contactType", contactType)
+      return modelAndView
```
petergphillips, line 56: looks like you are losing the `error` somewhere along the line?  Not passed through to the view?

petergphillips (dismissed):
LGTM

petergphillips (approved):
LGTM

PR #772, opened 2021-02-23
DT-1232: 🎨 Move user selector to own endpoint

petergphillips:


src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/resource/UserSelectorAuthorizationEndpoint.kt
```diff
@@ -41,40 +41,31 @@ class UserSelectorAuthorizationEndpoint(
...
-              ),
-              model
-            )
-          }
-        }
-        else -> {
-        } // no action required, will take the user to the user selector
       }
+      // take the user to select what user they would like to use
+      else -> ModelAndView("userSelector", model)
```
petergphillips, line 67: render view directly rather than go to another controller first

src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/resource/UserSelectorController.kt
```diff
@@ -2,17 +2,11 @@
...
+import org.springframework.web.bind.annotation.PostMapping
 
 @Controller
-@SessionAttributes("authorizationRequest")
 class UserSelectorController {
-  @GetMapping("/oauth/confirm_access")
-  fun userSelector(authentication: Authentication, authorizationRequest: AuthorizationRequest, model: Map<String, *>):
-    ModelAndView = ModelAndView("userSelector", model)
+  @PostMapping("/user-selector")
+  fun approveOrDeny(): String = "forward:/oauth/authorize"
```
petergphillips, line 11: at present just forward onto the `authorize` endpoint, but will be expanded later to take the user to a 2fa endpoint instead

JamieDraperUK (approved):
LGTM

PR #778, opened 2021-02-26
DT-1602: added token type for account details 2fa and added check that…

SimonMitchellMOJ:
… user has passed password step for change primary email


petergphillips (approved):
LGTM

PR #784, opened 2021-03-01
DT-1232: 🎨 Move MFA challenge post code to parent

petergphillips:


src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/resource/AbstractMfaController.kt
```diff
@@ -37,9 +41,9 @@ abstract class AbstractMfaController(
   }
 
   protected fun mfaChallengeRequestError(
-    @RequestParam error: String?,
-    @RequestParam token: String?,
-    @RequestParam mfaPreference: MfaPreferenceType?,
+    error: String?,
```
petergphillips, line 44: parent functions don't need the annotation

JamieDraperUK (approved):

PR #785, opened 2021-03-01
DT-1232: 🎨 Move MFA challenge code to parent

petergphillips:


src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/resource/MfaControllerAccountDetails.kt
```diff
@@ -42,25 +41,10 @@ class MfaControllerAccountDetails(
   fun mfaChallengeRequestAccountDetail(
     authentication: Authentication,
     @RequestParam contactType: String,
-    @RequestParam token: String?,
```
petergphillips, line 45: parameters found not to be used when moved to parent

src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/resource/MfaControllerAccountDetails.kt
```diff
@@ -72,7 +56,6 @@ class MfaControllerAccountDetails(
 
   @GetMapping("/account/mfa-challenge-error")
   fun mfaChallengeRequestAccountDetailError(
-    authentication: Authentication,
```
petergphillips, line 75: another parameter not used

JamieDraperUK (approved):
LGTM

PR #791, opened 2021-03-02
🚧 PPD-201: Matching NOMIS and Auth users by email

brightonsbox:
Work in progress. There is not yet any way of searching for
users across both Auth and NOMIS based on their email address.
This is work in progress as we have need to search potentially
by partial email address and first / last names too.

Because this pulls data from two datasets, this just returns
a list (no pagination).


src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/resource/api/UserController.kt
```diff
@@ -24,6 +26,35 @@ import java.security.Principal
...
+    value = [
+      ApiResponse(code = 200, message = "OK", response = UserDetail::class, responseContainer = "List"),
+      ApiResponse(code = 204, message = "No users found."),
+      ApiResponse(code = 400, message = "Bad request", response = ErrorDetail::class),
+      ApiResponse(code = 401, message = "Unauthorized", response = ErrorDetail::class)
+    ]
+  )
+  fun getUsers(
+    @ApiParam(value = "The email address of the user.") @RequestParam email: String?
+  ): ResponseEntity<Any> {
```
petergphillips, line 46: we're trying to move away from doing this and just returning the `List<UserDetails>` instead

src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/resource/api/UserController.kt
```diff
@@ -24,6 +26,35 @@ import java.security.Principal
...
+
+    if (email.isNullOrEmpty()) {
```
petergphillips, line 48: could be handled automatically instead by a `@NotEmpty` on the param?

src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/resource/api/UserController.kt
petergphillips: might need to always return a `200` rather than sometimes returning a `204` but normally mapping different statuses is harder work in the front end anyway...

src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/security/UserService.kt
```diff
@@ -173,6 +173,12 @@ class UserService(
         )
       }
   }
+
+  fun findUsersByEmail(email: String): List<UserPersonDetails> {
+    val authUsers = authUserService.findAuthUsersByEmail(email)
+    val nomisUsers = nomisUserService.getNomisUsersByEmail(email)
+    return (authUsers + nomisUsers).distinctBy { it.username }
```
petergphillips, line 180: this seems a bit odd to me - why search for external users and also nomis users?  I thought you just wanted nomis users?  You're missing out delius too?

petergphillips: I can't see the PPD-201 issue to understand what the requirements are though :disappointed:

timharrison-moj, line 180: The `(authUsers + nomisUsers).distinctBy` with select authUsers in preference to nomisUsers and just based on email address, so I think we might lose a lot of information this way? Maybe distinctBy { username, email } might be a better bet? Multiple accounts possibly have the same email address so we would probably want them all.

src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/resource/api/UserController.kt
```diff
@@ -24,6 +26,35 @@ import java.security.Principal
...
+      return ResponseEntity.badRequest()
+        .body(ErrorDetail(BAD_REQUEST.reasonPhrase, "No email address provided to search", "email"))
+    }
+
+    val users = userService.findUsersByEmail(email).map { UserDetail.fromPerson(it) }
+
```
timharrison-moj, line 54: Any plans to add firstname, lastname to the searc criteria? Or are you thinking different endpoints? They may even exist already.

src/test/kotlin/uk/gov/justice/digital/hmpps/oauth2server/security/UserServiceTest.kt
```diff
@@ -228,6 +228,54 @@ class UserServiceTest {
...
+      )
+
+      whenever(authUserService.findAuthUsersByEmail(anyString())).thenReturn(listOf(authUser))
+      whenever(nomisUserService.getNomisUsersByEmail(anyString())).thenReturn(listOf(nomisUser))
+
+      val found = userService.findUsersByEmail("some@email.com")
+
+      assertThat(found).hasSize(1)
+      assertThat(found[0]).isSameAs(authUser)
+    }
```
timharrison-moj, line 276: Maybe a test with multiple records containing the same email address to check that the distinctBy is selecting auth users in preference?

timharrison-moj (commented):
A couple of small comments & questions added.

src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/resource/api/UserController.kt
brightonsbox: Yep makes sense, updated.

src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/resource/api/UserController.kt
brightonsbox: Done

src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/security/UserService.kt
brightonsbox: Basic summary of requirements is that we need to search for user details based on their email address regardless of whether they are an auth user or a nomis user that just hasn't logged into DPS yet.  The idea here is that we can just call this one endpoint rather than calling two.

At the moment we are only expecting Auth and NOMIS users in our services, but I've added delius users here now too.

If this approach doesn't make sense please let me know.

brightonsbox: I've removed the distinctBy for now - will look to handle in the client side

src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/resource/api/UserController.kt
brightonsbox: I think this endpoint could be extended to consider this, but just want this functionality for now

src/test/kotlin/uk/gov/justice/digital/hmpps/oauth2server/security/UserServiceTest.kt
brightonsbox: now not needed

timharrison-moj (commented):
Generally looks good to me. Where does the role check occur for these endpoints? I didn't see a `@PreAuthorize("hasAnyRole('INTEL_ADMIN')")` or equivalent authority checking? Though these may be done somewhere outside these changes?  And do we want to restrict these endpoints to specific client roles?

PR #800, opened 2021-03-05
add new interventions auth user to sample data for local testing

tomsmyers:
this PR adds a new auth user and "INT_SP_" group to the sample data so we can test the interventions service locally without creating this group by hand.


petergphillips (approved):
LGTM

PR #802, opened 2021-03-08
DT-1656 update local seed data with new community API roles

tomsmyers:
the output of [this ticket](https://dsdmoj.atlassian.net/browse/DT-1656) was to remove the ROLE_AUTH_DELIUS_LDAP role and replace it with two more granular roles. this PR updates the local seed to data to use the new roles and allow users to log in with delius credentials.


petergphillips (approved):
LGTM

PR #805, opened 2021-03-11
DT-1739: ➖ Remove spring boot devtools from jar file

petergphillips:


build.gradle.kts
```diff
@@ -39,13 +39,13 @@ dependencies {
...
-  implementation("com.zaxxer:HikariCP:4.0.2")
+  implementation("org.flywaydb:flyway-core:7.7.0")
+  implementation("com.zaxxer:HikariCP:4.0.3")
   implementation("org.apache.commons:commons-text:1.9")
-  implementation("com.microsoft.sqlserver:mssql-jdbc:9.2.0.jre11")
+  implementation("com.microsoft.sqlserver:mssql-jdbc:9.2.1.jre11")
 
   runtimeOnly("com.h2database:h2:1.4.200")
-  runtimeOnly("org.springframework.boot:spring-boot-devtools")
+  developmentOnly("org.springframework.boot:spring-boot-devtools")
```
petergphillips, line 48: this is the recommended way to do this according to spring docs - see https://docs.spring.io/spring-boot/docs/current/reference/html/using-spring-boot.html#using-boot-devtools

mikehalmamoj (approved):

PR #806, opened 2021-03-11
AR-987 search auth users from multiple sources

timharrison-moj:
The Intelligence management  service, and possibly pin phones and UoF, need a way to search for potential users from a range of sources.  For IMS this is Nomis, Auth and Delius and for Pin Phones it's Nomis and Auth initially. 

The auth service already provided an endpoint which does most of this with `/auth/api/authuser/search` which matches a search term against username, email, first name and last name, and filters by status to provides a pageable result set.  However, this endpoint limited the user data only to AuthSource.auth in its UserFilter here :

`andBuilder.add(cb.equal(root.get<Any>("source"), AuthSource.auth))`

There are good reasons for this, as the role and group parameters are only applicable to auth users, but since the IMS and Pin phone requirements do not require the role and group filters it is still possible that this is a great fit. I have added a search endpoint into the UserController which does not provide the role or group filters, but does allow a list of auth sources to be passed in its query parameters. The existing behaviour of the /authuser/search endpoint is is maintained as it defaults to AuthSource.auth.

I've tested this locally and added unit/integration tests to verify both new and existing functionality and also checked that the auth sources, when passed through the user filter, produce a reasonable hibernate query that provides expected results.

I am aware of the limitations of this - it will only return user matches where they exist in auth - requiring that any users have first authenticated via auth to one of the DPS services, but I think this is a reasonable trade-off to re-use an existing endpoint.


src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/auth/model/UserFilter.kt
```diff
@@ -25,7 +26,12 @@ class UserFilter(
     cb: CriteriaBuilder
   ): Predicate {
     val andBuilder = ImmutableList.builder<Predicate>()
-    andBuilder.add(cb.equal(root.get<Any>("source"), AuthSource.auth))
+
+    if (!userSources.isNullOrEmpty()) {
```
petergphillips, line 30: wondering if this would be clearer in the `buildMultipleSourcesPredicate` i.e. have a `buildSourcesPredicate`

src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/auth/model/UserFilter.kt
```diff
@@ -52,6 +58,14 @@ class UserFilter(
     return cb.and(*andBuilder.build().toTypedArray())
   }
 
+  private fun buildMultipleSourcesPredicate(root: Root<User>, cb: CriteriaBuilder, sources: List<AuthSource>): Predicate {
+    val orBuilder = ImmutableList.builder<Predicate>()
+    for (source in sources) {
```
petergphillips, line 63: prefer `sources.forEach`

src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/maintain/AuthUserService.kt
```diff
@@ -133,14 +133,24 @@ class AuthUserService(
     searcher: String,
     authorities: Collection<GrantedAuthority>,
     status: UserFilter.Status,
+    userSources: List<AuthSource>?,
   ): Page<User> {
+
+    var incompatibleSources = false
```
petergphillips, line 139: prefer immutable variables

src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/maintain/AuthUserService.kt
```diff
@@ -133,14 +133,24 @@ class AuthUserService(
...
+    if (!userSources.isNullOrEmpty() && (!roleCodes.isNullOrEmpty() || !groupCodes.isNullOrEmpty())) {
+      incompatibleSources = userSources.any { it in listOf(AuthSource.nomis, AuthSource.delius, AuthSource.azuread) }
```
petergphillips, line 141: not really following this logic - isn't this effectively saying that `userSources` will automatically be set to `auth` if `roleCodes` or `groupCodes` is set?

src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/resource/api/AuthUserController.kt
```diff
@@ -139,19 +140,19 @@ class AuthUserController(
       value = "Sort column and direction, eg sort=lastName,desc"
     )
   )
-  @PreAuthorize("hasAnyRole('ROLE_MAINTAIN_OAUTH_USERS', 'ROLE_AUTH_GROUP_MANAGER')")
+  @PreAuthorize(
+    "hasAnyRole('ROLE_MAINTAIN_OAUTH_USERS', 'ROLE_AUTH_GROUP_MANAGER', 'ROLE_INTEL_ADMIN', 'ROLE_PPM_GLOBAL_ADMIN')"
```
petergphillips, line 144: like this, is clearer what roles can call this

petergphillips (commented):
Missing integration test changes

src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/auth/model/UserFilter.kt
timharrison-moj: Yes, agreed. I'll move it all there.

src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/maintain/AuthUserService.kt
timharrison-moj: Almost - it says if userSources is present and AND either groupCodes or roleCodes is is also present, check if  authSources contains any source other than auth, and if it does default the userSources back to auth. There might be a better way to express this more clearly so I'll have another go.

petergphillips: had another think about this - maybe we should actually throw a bad request here since the user has asked for something that is incompatible

src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/resource/api/AuthUserController.kt
```diff
@@ -139,19 +140,19 @@ class AuthUserController(
...
   fun searchForUser(
-    @ApiParam(
-      value = "The username, email or name of the user.",
-      example = "j smith"
-    ) @RequestParam(required = false) name: String?,
+    @ApiParam(value = "The username, email or name of the user.", example = "j smith") @RequestParam(required = false) name: String?,
     @ApiParam(value = "The role codes of the user.") @RequestParam(required = false) roles: List<String>?,
     @ApiParam(value = "The group codes of the user.") @RequestParam(required = false) groups: List<String>?,
     @ApiParam(value = "Limit to active / inactive / show all users.") @RequestParam(required = false, defaultValue = "ALL") status: Status,
+    @ApiParam(value = "Optional list of user sources [nomis|delius|auth|azuread] (incompatible with group and role filters)") @RequestParam(required = false) userSources: List<AuthSource>?,
```
petergphillips, line 151: might be better to be called _authentication sources_ rather than _user sources_

src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/maintain/AuthUserService.kt
timharrison-moj: Removed this and simplified the check.

src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/maintain/AuthUserService.kt
timharrison-moj: I've found a clearer way with this now.

timharrison-moj: I thought about the bad request option, and I put a comment about compatibility with roles/groups in the ApiParam text for user sources, but then I started thinking that if a client specifies roles or groups of course they meant to limit this to AuthSource.auth only! I don't mind either way, so happy to be guided on preference.

src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/maintain/AuthUserService.kt
```diff
@@ -133,14 +133,18 @@ class AuthUserService(
...
     authorities: Collection<GrantedAuthority>,
     status: UserFilter.Status,
+    authSources: List<AuthSource>?,
   ): Page<User> {
+    val sources = if (authSources.isNullOrEmpty()) listOf(AuthSource.auth) else authSources
     val groupSearchCodes = if (authorities.any { it.authority == "ROLE_MAINTAIN_OAUTH_USERS" }) {
       groupCodes
-    } else {
+    } else if (authorities.any { it.authority == "ROLE_AUTH_GROUP_MANAGER" }) {
       val assignableGroupCodes = authUserGroupService.getAssignableGroups(searcher, authorities).map { it.groupCode }
```
timharrison-moj, line 142: As there are now multiple potential roles calling the endpoint this needed to differentiate between the user maintainers, the group managers and others, like intel admin (or pin phone admin etc). where it did not need to look up the assignable roles, nor pass through the group codes.

src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/resource/api/UserController.kt
```diff
@@ -116,4 +128,116 @@ class UserController(private val userService: UserService) {
...
+        lastName = user.person?.lastName,
+        locked = user.locked,
+        enabled = user.isEnabled,
+        verified = user.verified,
+        lastLoggedIn = user.lastLoggedIn,
+        source = user.source,
+      )
+    }
+  }
 }
```
timharrison-moj, line 243: Some potential here to extract a base class for AuthUser and extend in both UserController and AuthUserController, but to avoid impacting too much existing functionality the AuthUserWithSource class is implemented in its entirety here.

src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/security/UserService.kt
```diff
@@ -173,6 +177,16 @@ class UserService(
...
+  fun searchUsersInMultipleSourceSystems(
+    name: String?,
+    pageable: Pageable,
+    searcher: String,
+    authorities: Collection<GrantedAuthority>,
+    status: UserFilter.Status,
+    authSources: List<AuthSource>?,
+  ): Page<User> =
+    authUserService.findAuthUsers(name, null, null, pageable, searcher, authorities, status, authSources)
 }
```
timharrison-moj, line 190: As the authUserService was already injected and used in the userService I made the call chain go from userController -> userService -> authUserService.

src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/auth/model/UserFilter.kt
timharrison-moj: Done.

src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/resource/api/AuthUserController.kt
timharrison-moj: Done - well its called `authSources`

brightonsbox (approved):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/maintain/AuthUserService.kt
```diff
@@ -133,14 +133,18 @@ class AuthUserService(
...
       if (groupCodes.isNullOrEmpty()) assignableGroupCodes else groupCodes.filter { g -> assignableGroupCodes.any { it == g } }
+    } else {
+      emptyList()
     }
-    val userFilter = UserFilter(name = name, roleCodes = roleCodes, groupCodes = groupSearchCodes, status)
+    val userFilter = UserFilter(name, roleCodes, groupCodes = groupSearchCodes, status, sources)
```
petergphillips, line 147: normally nicer in kotlin with multiple parameters in the constructor to call by name rather than rely on parameter order

src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/resource/api/AuthUserController.kt
```diff
@@ -139,19 +139,18 @@ class AuthUserController(
...
-    ) @RequestParam(required = false) name: String?,
+    @ApiParam(value = "The username, email or name of the user.", example = "j smith") @RequestParam(required = false) name: String?,
     @ApiParam(value = "The role codes of the user.") @RequestParam(required = false) roles: List<String>?,
     @ApiParam(value = "The group codes of the user.") @RequestParam(required = false) groups: List<String>?,
     @ApiParam(value = "Limit to active / inactive / show all users.") @RequestParam(required = false, defaultValue = "ALL") status: Status,
     @PageableDefault(sort = ["Person.lastName", "Person.firstName"], direction = Sort.Direction.ASC) pageable: Pageable,
     @ApiIgnore authentication: Authentication,
   ): Page<AuthUser> =
-    authUserService.findAuthUsers(name, roles, groups, pageable, authentication.name, authentication.authorities, status)
+    authUserService.findAuthUsers(name, roles, groups, pageable, authentication.name, authentication.authorities, status, null)
```
petergphillips, line 153: this looks naff, would be better to just add a default to the `findAuthUsers` method and then the field can also be not nullable

src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/resource/api/UserController.kt
```diff
@@ -116,4 +128,116 @@ class UserController(private val userService: UserService) {
 
   private fun notFoundResponse(username: String): ResponseEntity<Any?> = ResponseEntity.status(HttpStatus.NOT_FOUND)
     .body(ErrorDetail("Not Found", "Account for username $username not found", "username"))
+
+  @GetMapping("/api/user/search")
+  @ApiOperation(
+    value = "Search for users in the Auth DB who match on partial first name, surname, username or email and return a pageable result set. " +
+      "Optionally provide a list of authentication sources from any combination of [auth|delius|nomis|azuread] comma-separated." +
+      "Defaults to auth source only if not provided." +
+      "Will find users with an auth source of nomis, delius, auth and azuread if they have authenticated with auth at least once.",
```
petergphillips, line 137: Would be good to indicate that the data can be out of date.  Might be nicer to use a kotlin multiline string here, since it looks like some of the string concatenation will result in things like `provided.Will` as is

src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/resource/api/UserController.kt
```diff
@@ -116,4 +128,116 @@ class UserController(private val userService: UserService) {
 
   private fun notFoundResponse(username: String): ResponseEntity<Any?> = ResponseEntity.status(HttpStatus.NOT_FOUND)
     .body(ErrorDetail("Not Found", "Account for username $username not found", "username"))
+
+  @GetMapping("/api/user/search")
+  @ApiOperation(
+    value = "Search for users in the Auth DB who match on partial first name, surname, username or email and return a pageable result set. " +
+      "Optionally provide a list of authentication sources from any combination of [auth|delius|nomis|azuread] comma-separated." +
```
petergphillips, line 135: don't think the comma separated stuff here is true?  Think it should actually be that if you want multiple then you need to specify the parameter multiple times?

src/test/kotlin/uk/gov/justice/digital/hmpps/oauth2server/resource/api/AuthUserControllerTest.kt
```diff
@@ -771,7 +772,7 @@ class AuthUserControllerTest {
   @Test
   fun searchForUser() {
     val unpaged = Pageable.unpaged()
-    whenever(authUserService.findAuthUsers(anyString(), anyList(), anyList(), any(), anyString(), any(), any())).thenReturn(
+    whenever(authUserService.findAuthUsers(anyString(), anyList(), anyList(), any(), anyString(), any(), any(), anyOrNull())).thenReturn(
```
petergphillips, line 775: if the parameter is changed to have a default anyway then this doesn't need to change

petergphillips (dismissed):
Approved with minor comments

src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/resource/api/UserController.kt
timharrison-moj: Calling this would be `/api/user/search?search=xxxx&authSources=nomis&authSources=delius` -  I'll adjust the wording to reflect this.

src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/resource/api/UserController.kt
timharrison-moj: I've updated this to a multiline string and added words around the potential for data in the auth DB to be out of date with the users' source system data.

src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/maintain/AuthUserService.kt
timharrison-moj: Ok will reinstate the parameter names here.

src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/resource/api/AuthUserController.kt
timharrison-moj: I've defaulted this in findAuthUsers now.

src/test/kotlin/uk/gov/justice/digital/hmpps/oauth2server/resource/api/AuthUserControllerTest.kt
timharrison-moj: Done.

petergphillips (approved):
LGTM

PR #808, opened 2021-03-12
Gd/remove soct roles from mfa list

garethmdavies:
This PR will remove the Pathfinder and Pin phone roles from the MFA list as both services will now use the service based MFA managed by the HMPPS Auth client configuration.


petergphillips (approved):
LGTM

PR #816, opened 2021-03-17
DT-1747 Sanitise logging data by removing crlf

karenmillermoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/utils/StringSanitiser.kt
```diff
@@ -0,0 +1,7 @@
+package uk.gov.justice.digital.hmpps.oauth2server.utils
+
+fun String.sanitise(): String {
+  return this
```
petergphillips, line 4: don't need the `return` - can be a single expression function

petergphillips: intellij should hopefully come up with the expression body suggestion

src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/resource/api/AuthUserController.kt
```diff
@@ -419,15 +415,16 @@ class AuthUserController(
...
+      log.info("Amend user succeeded for user $username".sanitise())
       if (smokeTestEnabled) {
         ResponseEntity.ok(resetLink)
       } else ResponseEntity.noContent().build()
     } catch (e: EntityNotFoundException) {
       ResponseEntity.status(HttpStatus.NOT_FOUND)
         .body(ErrorDetail("username.notfound", "User not found", "username"))
     } catch (e: VerifyEmailException) {
-      log.info("Amend user failed for user {} for field email with reason {}", username, e.reason)
+      val fred = "something\r\nweird"
```
petergphillips, line 426: `fred`?

src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/resource/ResetPasswordController.kt
```diff
@@ -85,7 +86,7 @@ class ResetPasswordController(
           modelAndView.addObject("resetLink", resetLink.get())
         }
       } else {
-        log.info("Reset password request failed, no link provided for {}", usernameOrEmail)
+        log.info("Reset password request failed, no link provided for $usernameOrEmail".sanitise())
```
petergphillips, line 89: difficult one.  Normally with loggers we pass through the parameters to the `log` statement so that it is done when the log item is actually logged, also only really need to sanitise the `usernameOrEmail` i.e.
```
log.info("Reset password request failed, no link provided for {}", usernameOrEmail.sanitise())
```
might be better, however maybe safer to sanitise the whole lot :confused:

src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/utils/StringSanitiser.kt
```diff
@@ -0,0 +1,7 @@
+package uk.gov.justice.digital.hmpps.oauth2server.utils
+
+fun String.sanitise(): String {
```
petergphillips, line 3: not a great fan of this name, doesn't really explain what it is doing.  Maybe `removeAllLfCr` instead would be clearer?

src/test/kotlin/uk/gov/justice/digital/hmpps/oauth2server/utils/StringSanitiserTest.kt
```diff
@@ -0,0 +1,21 @@
...
+
+internal class StringSanitiserTest {
+  @Test
+  fun `carriage return is sanitised`() {
+    assertThat(("SomeUser\rWithHiddenCR").sanitise()).isEqualTo("SomeUserWithHiddenCR")
+  }
+
+  @Test
+  fun `linefeed is sanitised`() {
+    assertThat(("SomeUser\nWithHiddenCR").sanitise()).isEqualTo("SomeUserWithHiddenCR")
```
petergphillips, line 14: would be good to test that multiple CRs / LFs are removed, maybe the `sanitise` method just removes the first one of each?

karenmillermoj: Good idea - will add in

src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/resource/api/AuthUserController.kt
karenmillermoj: good spot - what on earth is that!

src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/utils/StringSanitiser.kt
karenmillermoj: Fab - Updated

src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/utils/StringSanitiser.kt
karenmillermoj: Yeah - I wasn't sure whether to be more specific, expected possibly further things to add to it.  I suppose if further things are added, we can change the name at that point.  Updated.

src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/resource/ResetPasswordController.kt
karenmillermoj: Yeah - difficult to decide which is best.  It is only the user input that needed to be sanitised.  Shouldn't affect anything.  Have left as is for now.

andymarke (approved):
LGTM

PR #819, opened 2021-03-23
AR-1119: Persist first and last names after login for Nomis and Delius auth sources

timharrison-moj:
For Nomis users -> Persists the first and last names from the staff details provided in the NomisUserPersonDetails into the Auth user table after a successful login.  Email address is already populated but only on first login, and subsequent logins do not update it.

For Delius users -> we already update the email address & verified flag so this PR also persists the first and last names from the DeliusUserPerson details into the Auth DB after a successful login.

For AzureAD users - no change, the email, first and last name are already persisted.
For Auth users - no change, the email, first and last names are already persisted.

This will enable the multi-source search endpoint to match with first & last names for all sources and provide full results.


petergphillips (approved):
LGTM

PR #822, opened 2021-03-24
✨ PPD-224: Adding new endpoint to retrieve user roles (Nomis/Auth/Delius)

brightonsbox:


src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/resource/api/UserController.kt
```diff
@@ -96,6 +97,31 @@ class UserController(private val userService: UserService) {
       .map { ResponseEntity.ok(it) }.orElse(notFoundResponse(username))
   }
 
+  @GetMapping("/api/user/{username}/roles")
+  @ApiOperation(
+    value = "List of roles for user.",
```
petergphillips, line 102: would be good to document what roles are required in the api docs too

src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/resource/api/UserController.kt
```diff
@@ -96,6 +97,31 @@ class UserController(private val userService: UserService) {
...
+    notes = "List of roles for user.",
+    nickname = "userRoles",
+    consumes = "application/json",
+    produces = "application/json"
+  )
+  @ApiResponses(
+    value = [
+      ApiResponse(code = 200, message = "OK", response = UserRole::class, responseContainer = "List"),
```
petergphillips, line 110: don't need the 200 response here, will be generated automatically

petergphillips: although would be good to document the 404

src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/resource/api/UserController.kt
brightonsbox: 👍

petergphillips (approved):
LGTM

PR #823, opened 2021-03-25
DT-1805: ✨ First stab at jpa version of client details

petergphillips:


src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/auth/model/Client.kt
```diff
@@ -0,0 +1,68 @@
...
+import javax.persistence.Id
+import javax.persistence.Table
+
+@Suppress("JpaDataSourceORMInspection")
+@Entity
+@Table(name = "oauth_client_details")
+data class Client(
+  @Id
+  @Column(name = "client_id", nullable = false)
+  val id: String
```
petergphillips, line 18: will map more in future, but just bare minimum at present

SimonMitchellMOJ (approved):
LGTM

PR #830, opened 2021-03-30
🗃 PPD-210: Create PPM roles in auth

brightonsbox:


JamieDraperUK (approved):

src/main/resources/db/auth/V5_28__add_ppm_user_roles.sql
```diff
@@ -0,0 +1,4 @@
+INSERT INTO roles (role_id, role_code, role_name)
+VALUES (newid(), 'PPM_ANALYST',             'PPM Analyst'),
+       (newid(), 'PPM_AUTHORISING_OFFICER', 'PPM Authorising Officer'),
+       (newid(), 'PPM_GLOBAL_ADMIN',        'PPM Global Admin');
```
petergphillips, line 4: Normally any insert into `roles` should also be associated with an insert into `GROUP_ROLES` so that group managers can administer the roles, unless these roles should only ever be assigned by admins?

PR #831, opened 2021-03-31
P4-2724 new PECS role for the journey price calculation service for maintenance of journey prices.

matt-crowson-moj:
**Changes:**

This purpose of the role is to enable us to control which users can manage journey prices in the PECS Journey Price Calculation service.


src/main/resources/db/auth/V5_29__add_pecs_price_maintenance_role.sql
```diff
@@ -0,0 +1,2 @@
+INSERT INTO roles (role_id, role_code, role_name)
+VALUES (newid(), 'PECS_MAINTAIN_PRICE', 'PECS Journey Price Maintenance');
```
petergphillips, line 2: normally any insert into `ROLES` table is associated with an insert into the `GROUP_ROLES` table otherwise group managers can't administer the role.  Should this role ever be added by group managers?

petergphillips (approved):
LGTM

PR #834, opened 2021-04-06
DT-1807: ✨ Track client last accessed

petergphillips:


src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/service/ClientService.kt
```diff
@@ -28,8 +28,8 @@ class ClientService(
     clientRepository.findAll().sortedBy { it.id }.distinctBy { baseClientId(it.id) }
 
   fun loadClientWithCopies(clientId: String): ClientDetailsWithCopies {
-    val clients = find(clientId).filter { it.id != clientId }
```
petergphillips, line 31: show all matching clients, even current one so that get view of last accessed for all

src/main/resources/db/auth/V5_30__add_client_details_last_accessed.sql
```diff
@@ -0,0 +1,2 @@
+alter table oauth_client_details
+    add last_accessed DATETIME2 NOT NULL DEFAULT CURRENT_TIMESTAMP;
```
petergphillips, line 2: default new items and all existing to current date time

JamieDraperUK (approved):
LGTM

PR #837, opened 2021-04-08
DT-1808: clientId and clientSecret in base64 displayed on confirmation…

SimonMitchellMOJ:
… page if clientSecret is added/updated


petergphillips (approved):
LGTM

PR #842, opened 2021-04-12
DT-1539: ✨ Add uuid to user me endpoint

petergphillips:


src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/resource/api/UserController.kt
```diff
@@ -54,8 +54,11 @@ class UserController(private val userService: UserService) {
     ]
   )
   fun me(@ApiIgnore principal: Principal): UserDetail {
-    val user = userService.findMasterUserPersonDetails(principal.name)
-    return user.map { UserDetail.fromPerson(it) }.orElse(UserDetail.fromUsername(principal.name))
+    val upd = userService.findMasterUserPersonDetails(principal.name)
+    return upd.map {
+      val user = userService.getOrCreateUser(principal.name).orElseThrow()
```
petergphillips, line 59: we have to ensure that the user exists in auth in order to then get a uuid

SimonMitchellMOJ (approved):
LGTM

PR #846, opened 2021-04-14
DT-1642: ✨ Add create group

petergphillips:


src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/maintain/GroupsService.kt
```diff
@@ -146,6 +147,22 @@ class GroupsService(
     )
   }
 
+  @Transactional(transactionManager = "authTransactionManager")
+  @Throws(GroupExistsException::class, GroupNotFoundException::class)
```
andymarke, line 151: why do you need the throws annotation? Is it because we still have Java calling these functions?

andrewrlee (approved):
LGTM

andymarke (approved):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/maintain/GroupsService.kt
SimonMitchellMOJ, line 151: Will we ever have a GroupNotFoundExecption?

petergphillips: good question, will see if I can remove it and everything still works :grin:

petergphillips: good spot, will fix in next PR

petergphillips: test failed
```
org.mockito.exceptions.base.MockitoException: 
Checked exception is invalid for this method!
```
wouldn't allow me to throw the exception in the test, but guess that is a test issue

PR #852, opened 2021-04-19
DT-1900: added ip allowlist protection to /api endpoints

SimonMitchellMOJ:


src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/config/AuthenticationManagerConfiguration.kt
```diff
@@ -77,6 +77,7 @@ class AuthenticationManagerConfiguration(
       .antMatchers(HttpMethod.GET, "/login").permitAll()
       .antMatchers(HttpMethod.POST, "/login").permitAll()
       .antMatchers("/ui/**").access("isAuthenticated() and @authIpSecurity.check(request)")
+      .antMatchers("/api/**").access("isAuthenticated() and @authIpSecurity.check(request)")
```
petergphillips, line 80: think this should be `/api/client`?

petergphillips (approved):
LGTM

PR #860, opened 2021-04-23
DT-1951: delete client deployment info when client is removed and no duplicates exist

SimonMitchellMOJ:


src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/auth/repository/ClientDeploymentRepository.kt
```diff
@@ -6,4 +6,6 @@ import uk.gov.justice.digital.hmpps.oauth2server.auth.model.ClientDeployment
 interface ClientDeploymentRepository : CrudRepository<ClientDeployment, String> {
 
   fun findByBaseClientId(baseClientId: String): ClientDeployment?
+
+  fun deleteByBaseClientId(baseClientId: String)
```
petergphillips, line 10: Is this function required - won't it be provided automatically by the `CrudRepository`

petergphillips: actually same with the `find` method, should just be able to use `findByIdOrNull`

petergphillips (approved):
LGTM

PR #866, opened 2021-04-28
DT-1962: ✨ Refactor prior to adding in caching

petergphillips:


src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/config/JWTTokenEnhancer.kt
```diff
@@ -17,7 +17,7 @@ import java.util.Optional
 
 class JWTTokenEnhancer : TokenEnhancer {
   @Autowired
-  private lateinit var clientsDetailsService: JdbcClientDetailsService
+  private lateinit var clientsDetailsService: ClientDetailsService
```
petergphillips, line 20: switch to interface

src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/resource/ClientsController.kt
```diff
@@ -6,8 +6,8 @@ import com.microsoft.applicationinsights.TelemetryClient
 import org.springframework.security.access.prepost.PreAuthorize
 import org.springframework.security.core.Authentication
 import org.springframework.security.core.GrantedAuthority
+import org.springframework.security.oauth2.provider.ClientRegistrationService
```
petergphillips, line 9: switch to interface for updates - different interface compared to reads

JamieDraperUK (approved):
LGTM

PR #874, opened 2021-05-04
DT-1986: updated redirects on add client success page generate client…

SimonMitchellMOJ:
… secret and client deployment details page to take user back to the client details page

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/resource/ClientsController.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/resource/ClientsController.kt
@@ -75,7 +75,7 @@ class ClientsController(
 
     clientService.saveClientDeploymentDetails(clientDeployment)
 
-    return ModelAndView("redirect:/ui/clients/form?client=${clientDeployment.baseClientId}")
+    return ModelAndView("redirect:/ui/clients/form", "client", clientDeployment.baseClientId)
   }
 
   @PostMapping("/add")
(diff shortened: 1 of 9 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/resource/ClientsController.kt
```diff
@@ -74,7 +75,7 @@ class ClientsController(
 
     clientService.saveClientDeploymentDetails(clientDeployment)
 
-    return ModelAndView("redirect:/ui")
+    return ModelAndView("redirect:/ui/clients/form?client=${clientDeployment.baseClientId}")
```
petergphillips, line 78: should really `addObject` to the `ModelAndView` rather than manually adding in the parameters here

src/main/resources/templates/ui/clientSuccess.html
```diff
@@ -18,7 +19,8 @@ <h1 class="govuk-panel__title"  th:text="${#strings.equals(newClient, 'true')} ?
   <dl class="govuk-summary-list">
 
     <div class="govuk-summary-list__row">
-      <dt class="govuk-summary-list__key" th:text="${#strings.equals(newClient, 'true')} ? 'new clientId' : 'clientId'">>
+      <dt class="govuk-summary-list__key" th:text="${#strings.equals(newClient, 'true')} ? 'new clientId' : 'clientId'">
+        >
```
petergphillips, line 23: is this an extra `>` by mistake?

src/main/resources/templates/ui/clientSuccess.html
```diff
@@ -53,8 +55,7 @@ <h1 class="govuk-panel__title"  th:text="${#strings.equals(newClient, 'true')} ?
     </div>
   </dl>
 
-  <a class="govuk-button" th:href="@{/ui}" id="continue">Continue</a>
-</div>
```
petergphillips, line 57: lost a `<div>` here, or was it an extra one by mistake?

petergphillips (dismissed):
LGTM

src/main/resources/templates/ui/clientSuccess.html
SimonMitchellMOJ: it was an extra one - so un-needed

petergphillips (approved):
LGTM

PR #888, opened 2021-05-18
DT-1394: Updated automatic deactivation reason message to use number …

SimonMitchellMOJ:
…of days from properties file instead of being hard coded


petergphillips (approved):
LGTM

PR #891, opened 2021-05-18
IC-1739 add interventions 'contract reference' groups to local dev data

tomsmyers:
This change updates the local test data to include some interventions contract groups, which allow for accessing interventions with contract scoped access restrictions when using the local dev instance of the interventions service.


Mjwillis (approved):
LGTM

PR #892, opened 2021-05-21
DT-2062: 🐛️ updated so when user has email as username and adds secondary…

SimonMitchellMOJ:
… email method does not error before verifyEmail email sent page


petergphillips (approved):
LGTM

PR #902, opened 2021-06-04
DCS-887 Create a 'read only' HDC RO role for Delius users

bell-pa:
New granted authority 'ROLE_LICENCE_RO_VIEW_ONLY' added to Delius role mappings for Delius role LHDCBT001.


GurnankCheema (approved):
LGTM

PR #905, opened 2021-06-07
update local docker oauth2 config for interventions to use assess risks and needs

tomsmyers:
this PR makes some small changes to the client config for 'interventions' in the local docker.

a couple of new roles are added and the redirect urls are updated to fix our logout screen.


petergphillips (approved):
LGTM

PR #907, opened 2021-06-07
Add CRS provider role as a default to all CRS provider groups

sldblog:
Default groups and roles for the dynamic framework (now called commissioned rehabilitative services) providers.

- Adds default role (`ROLE_CRS_PROVIDER`)
- Adds the role to all provider groups (previously applied manually) as a default role

### DIFF
--- a/src/main/resources/db/auth/V5_39__add_crs_role.sql
+++ b/src/main/resources/db/auth/V5_39__add_crs_role.sql
@@ -1,0 +1,7 @@
+INSERT INTO roles (role_id, role_code, role_name)
+VALUES (newid(), 'CRS_PROVIDER', 'Commissioned rehabilitative services (CRS) provider');
+
+INSERT INTO group_assignable_role (role_id, group_id, automatic)
+  SELECT role_id, g.group_id, 1
+  FROM roles r, groups g
+  WHERE r.role_code = 'CRS_PROVIDER' AND g.group_code LIKE 'INT_SP_%';


sldblog:
Hmm, it looks like the upsert broke some tests -- I'm assuming it's due to differences in MS SQL and the in-memory DB. I'll remove that migration

src/main/resources/db/auth/V5_38__add_crs_role.sql
```diff
@@ -0,0 +1,7 @@
+INSERT INTO roles (role_id, role_code, role_name)
```
petergphillips, line 1: 5_38 now taken I'm afraid, will have to be 5_39

src/main/resources/db/auth/V5_38__add_crs_role.sql
```diff
@@ -0,0 +1,7 @@
...
+VALUES (newid(), 'CRS_PROVIDER', 'Commissioned rehabilitative services (CRS) provider');
+
+INSERT INTO group_assignable_role (role_id, group_id, automatic)
+  SELECT role_id, g.group_id, 1
+  FROM roles r, groups g
+  WHERE r.role_code = 'CRS_PROVIDER' AND g.group_code LIKE 'INT_SP_%';
```
petergphillips, line 7: changes look fine, would have thought that you would want to insert into `USER_ROLE` for all the relevant users too though?

sldblog: there aren't any users yet as far as I know so that _should_ be all right (famous last words)

src/main/resources/db/auth/V5_38__add_crs_role.sql
sldblog: done in 8eed2c5774062417a81ddeacf43255d3090b1ded

petergphillips (approved):
LGTM

sldblog:
🙇‍♂️ Thank you!

PR #912, opened 2021-06-14
DT-2155 ensure locked out nomis user can reset their password

karenmillermoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/security/UserService.kt
```diff
@@ -47,9 +48,9 @@ class UserService(
       .or { azureUserService.getAzureUserByUsername(username).map { UserPersonDetails::class.java.cast(it) } }
       .or { deliusUserService.getDeliusUserByUsername(username).map { UserPersonDetails::class.java.cast(it) } }
 
-  fun findEnabledMasterUserPersonDetails(username: String): UserPersonDetails? =
+  fun findEnabledOrNomisLockedUserPersonDetails(username: String): UserPersonDetails? =
     authUserService.getAuthUserByUsername(username).filter { it.isEnabled }.map { UserPersonDetails::class.java.cast(it) }
-      .or { nomisUserService.getNomisUserByUsername(username).filter { it.isEnabled }.map { UserPersonDetails::class.java.cast(it) } }
+      .or { nomisUserService.getNomisUserByUsername(username).filter { it.isEnabled || it.accountDetail.status == AccountStatus.LOCKED }.map { UserPersonDetails::class.java.cast(it) } }
```
karenmillermoj, line 53: The filtering for a nomis user does end up returning the same as not filtering at all, as it covers all the different AccountStatus values.  Have left as it is as it is more explicit, but obv not needed.

petergphillips (approved):
LGTM

PR #913, opened 2021-06-15
DT-2171: 🔥 Remove group manager from mfa list as manage auth accounts…

petergphillips:
… now protected instead


SimonMitchellMOJ (approved):
LGTM

PR #922, opened 2021-06-24
PUD-89: Add additional callback urls for ppud-ui-client

philsegal:
- also ensured ppud_user is verified.


petergphillips (approved):
LGTM

PR #923, opened 2021-06-25
DT-2174: 🐛 Increase authorities size and tidy field on maintenance page

petergphillips:


src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/auth/model/Client.kt
```diff
@@ -58,9 +58,8 @@ data class Client(
     lastAccessed = LocalDateTime.now()
   }
 
-  fun resetSecretUpdated() {
```
petergphillips, line 61: unused

src/main/resources/templates/ui/form.html
```diff
@@ -220,8 +220,8 @@ <h2 class="govuk-heading-m">Client details</h2>
     </div>
     <div class="govuk-form-group">
       <label class="govuk-label" for="authorities">Authorities</label>
-      <input class="govuk-input" id="authorities" type="text" name="authorities"
-             th:value="${#strings.setJoin(clientDetails.authorities, ',')}"/>
+      <textarea class="govuk-textarea" id="authorities" type="text" name="authoritiesWithNewlines" rows="6"
```
petergphillips, line 223: make into textarea

src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/resource/ClientsController.kt
```diff
@@ -193,6 +202,17 @@ class ClientsController(
       super.setAuthorities(rolePrefixedAuthorities)
     }
 
+    // used by thymeleaf in form.html
```
andymarke, line 205: Would this work ? with commas and without?

ROLE_COMMUNITY,
ROLE_ANOTHER,
ROLE_ANDANOTHER
ROLE_LASTONE

andymarke (approved):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/resource/ClientsController.kt
petergphillips: yes, since all the whitespace (incl. newlines) is converted into commas before then being split on the commas and nulls from the resultant list removed

PR #924, opened 2021-06-25
DT-2174: 🐛 Change redirects to be textarea instead

petergphillips:


src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/resource/ClientsController.kt
```diff
@@ -205,13 +205,22 @@ class ClientsController(
     // used by thymeleaf in form.html
     var authoritiesWithNewlines: String?
       get() = authorities?.map { it.authority.substringAfter("ROLE_") }?.joinToString("\n")
-      set(authorisedRolesWithNewlines: String?) {
+      set(authorisedRolesWithNewlines) {
```
petergphillips, line 208: not needed as same type as `var`

src/main/resources/templates/ui/form.html
```diff
@@ -221,7 +221,7 @@ <h2 class="govuk-heading-m">Client details</h2>
     <div class="govuk-form-group">
       <label class="govuk-label" for="authorities">Authorities</label>
       <textarea class="govuk-textarea" id="authorities" type="text" name="authoritiesWithNewlines" rows="6"
-                th:text="*{clientDetails.authoritiesWithNewlines}"></textarea>
+                th:text="*{clientDetails.authoritiesWithNewlines}">an authority</textarea>
```
petergphillips, line 224: just prevent intellij warning about blank tag

mikehalmamoj (approved):

PR #926, opened 2021-06-28
DT-2129: validate email address supplied for forgotten username

SimonMitchellMOJ:


build.gradle.kts
```diff
@@ -1,8 +1,8 @@
 plugins {
   id("uk.gov.justice.hmpps.gradle-spring-boot") version "3.3.0"
-  kotlin("plugin.spring") version "1.5.10"
-  kotlin("plugin.jpa") version "1.5.10"
-  id("org.jlleitschuh.gradle.ktlint") version "10.0.0"
+  kotlin("plugin.spring") version "1.5.20"
```
petergphillips, line 3: shouldn't really upgrade this unless actually upgrading kotlin - could try `3.3.1-beta`

build.gradle.kts
```diff
@@ -1,8 +1,8 @@
...
+  kotlin("plugin.jpa") version "1.5.20"
+  id("org.jlleitschuh.gradle.ktlint") version "10.1.0"
```
petergphillips, line 5: I think this is in the plugin these days too?

src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/resource/ForgottenUsernameController.kt
```diff
@@ -60,6 +65,14 @@ class ForgottenUsernameController(
         }
       }
       modelAndView
+    } catch (e: ValidEmailException) {
+      log.error("Failed to send forgotten username email due to", e)
```
petergphillips, line 69: this log message doesn't look right

src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/resource/ForgottenUsernameController.kt
```diff
@@ -60,6 +65,14 @@ class ForgottenUsernameController(
...
+      telemetryClient.trackEvent(
+        "ForgottenUsernameRequestFailure",
```
petergphillips, line 71: don't we prefix all our events with `Auth`?

src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/resource/ForgottenUsernameController.kt
```diff
@@ -60,10 +65,18 @@ class ForgottenUsernameController(
         }
       }
       modelAndView
+    } catch (e: ValidEmailException) {
+      log.error("Forgotten username email failed validation")
```
petergphillips, line 69: bit naff this is an `error` - it is caused by the client entering invalid data so really could be an `info`, or don't need it at all...

petergphillips (approved):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/resource/ForgottenUsernameController.kt
SimonMitchellMOJ: will remove as don't think we need it really

petergphillips (approved):
LGTM

PR #929, opened 2021-06-30
DT-2097: schedule pre disable warning email to be sent 7 days before …

SimonMitchellMOJ:
…user account disabled for inactivity

### DIFF
--- a/helm_deploy/values-prod.yaml
+++ b/helm_deploy/values-prod.yaml
@@ -47,6 +47,7 @@ mfa_allowlist:
 
 env:
   SPRING_PROFILES_ACTIVE: "oracle,delius,token-verification"
+  APPLICATION_SIGNIN_URL: "https://sign-in.hmpps.service.justice.gov.uk/auth/login"
   APPLICATION_SUPPORT_URL: "https://support.hmpps.service.justice.gov.uk/feedback-and-support"
   APPLICATION_AUTHENTICATION_MATCH_SUBDOMAINS: "false"
   APPLICATION_AUTHENTICATION_UI_ALLOWLIST: "81.134.202.29,217.33.148.210"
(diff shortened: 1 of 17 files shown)


helm_deploy/values-prod.yaml
```diff
@@ -47,6 +47,7 @@ mfa_allowlist:
 
 env:
   SPRING_PROFILES_ACTIVE: "oracle,delius,token-verification"
+  APPLICATION_SIGNIN_URL: "https://sign-in.hmpps.service.justice.gov.uk/auth/login"
```
petergphillips, line 50: I'd be inclined to just have this as `/auth/` instead, they'll automatically be directed to login when going to that url

src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/timed/NotifyPreDisableAuthUsersService.kt
```diff
@@ -0,0 +1,84 @@
...
+  @Transactional(transactionManager = "authTransactionManager")
+  override fun processInBatches(): Int {
+    val usersToNotifyPreDisable =
+      repository.findByLastLoggedInBeforeAndEnabledIsTrueAndPreDisableWarningIsFalseAndSourceOrderByLastLoggedIn(
+        LocalDateTime.now().minusDays(loginDays.toLong()).minusHours(1L),
+        AuthSource.auth,
+      )
+    usersToNotifyPreDisable.forEach(
+      Consumer { user: User ->
+        if (user.verified) {
```
petergphillips, line 40: this looks weird - shouldn't it be on the query instead?

petergphillips: actually fine, just seen `else` block

petergphillips (dismissed):
LGTM

petergphillips (approved):
LGTM

PR #934, opened 2021-07-05
DT-2287: 🐛 Try credentials twice - second time after decoding

petergphillips:


build.gradle.kts
```diff
@@ -4,6 +4,8 @@ plugins {
   kotlin("plugin.jpa") version "1.5.20"
 }
 
+extra["spring-security.version"] = "5.5.1"
```
petergphillips, line 7: override downgrade in gradle plugin as can now use latest

src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/config/UrlDecodingRetryDaoAuthenticationProvider.kt
```diff
@@ -36,4 +40,23 @@ class LoggingDaoAuthenticationProvider(
...
+    userDetails: UserDetails?,
+    authentication: UsernamePasswordAuthenticationToken?
+  ) {
+    try {
+      super.additionalAuthenticationChecks(userDetails, authentication)
+    } catch (e: BadCredentialsException) {
+      authentication?.credentials?.let {
+        val decodedAuthentication = UsernamePasswordAuthenticationToken(
+          authentication.principal,
+          URLDecoder.decode(authentication.credentials.toString(), StandardCharsets.UTF_8.toString()),
```
petergphillips, line 54: try again with decoding of client secret.  No point decoding client id as we don't use any special characters in that

mikehalmamoj (approved):

PR #938, opened 2021-07-06
Revert "DT-2097: updated notify pre-disable schedule to 30 mins"

SimonMitchellMOJ:
Reverts ministryofjustice/hmpps-auth#937


petergphillips (approved):
LGTM

PR #939, opened 2021-07-06
DT-2287: ⬆️ Upgrade spring security

petergphillips:


build.gradle.kts
```diff
@@ -1,11 +1,9 @@
 plugins {
-  id("uk.gov.justice.hmpps.gradle-spring-boot") version "3.3.3"
+  id("uk.gov.justice.hmpps.gradle-spring-boot") version "3.3.4"
   kotlin("plugin.spring") version "1.5.20"
   kotlin("plugin.jpa") version "1.5.20"
 }
 
-extra["spring-security.version"] = "5.5.1"
```
petergphillips, line 7: don't need to pin any more as gradle plugin provides the correct version

SimonMitchellMOJ (approved):
LGTM

PR #943, opened 2021-07-08
DT-2304: ✨ Add create secret options to rotate script

petergphillips:


rotate_clientID_cloudplatform_app.sh
```diff
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 set -e
 
-if ! echo "$BASH_VERSION" | grep -E "^[45]" &> /dev/null; then
+if ! echo "$BASH_VERSION" | grep -E "^[45]" &>/dev/null; then
```
petergphillips, line 4: just reformatted the whole file using intellij

rotate_clientID_cloudplatform_app.sh
```diff
@@ -69,28 +96,27 @@ secretName=$(echo "${clientInfo_json}" | jq -r .clientDeployment.secretName)
 clientIdKey=$(echo "${clientInfo_json}" | jq -r .clientDeployment.clientIdKey)
 secretKey=$(echo "${clientInfo_json}" | jq -r .clientDeployment.secretKey)
 
-# Duplicate clientID get new secret
-results_json=$(hmpps_auth PUT "${HOST}/auth/api/client/${BASE_CLIENT_ID}")
```
petergphillips, line 73: move this after the secret stuff.  If the secret stuff fails then there is no point creating a new secret

rotate_clientID_cloudplatform_app.sh
```diff
@@ -13,9 +13,36 @@ fi
...
+while getopts ":cke:u:n" arg; do
+  case "${arg}" in
+  e) # Environment to run against
+    ENV="${OPTARG}"
+    ;;
+  u) # User to run script against
+    USER="${OPTARG}"
+    ;;
+  c) # Specify whether to create the kubernetes secret
+    create_secret=true
```
mattops, line 30: should this default to false? Just concerned that if incorrect deployment details are provided we may end up creating lots of bad k8s secrets.  But then I guess the deployment will fail right when it's restarted, right?

mattops: Ignore this. I see what's happening now.

rotate_clientID_cloudplatform_app.sh
```diff
@@ -69,28 +96,27 @@ secretName=$(echo "${clientInfo_json}" | jq -r .clientDeployment.secretName)
...
-currentClientID=$(kubectl -n "${namespace}" get secrets "${secretName}" -o json | jq -r ".data[\"${clientIdKey}\"] | @base64d")
-
 # Check if $secretKey exists and is readable
 if ! kubectl -n "${namespace}" get secrets "${secretName}" -o json | jq -e "select(.data[\"${secretKey}\"] != null)" &>/dev/null; then
   echo "Unable to find k8s secret with key \"${secretKey}\" in \"${secretName}\" for namespace \"${namespace}\""
-  exit 1
+  [[ -z "${create_keys_in_secret}" ]] && exit 1
 fi
 
 # Check if $deployment exists and is readable
```
mattops, line 122: I wonder if we should check the deployment first? near line 98

rotate_clientID_cloudplatform_app.sh
```diff
@@ -69,28 +96,27 @@ secretName=$(echo "${clientInfo_json}" | jq -r .clientDeployment.secretName)
...
+    exit 1
+  fi
+fi
 
 # Check if $clientIdKey exists and is readable
 if ! kubectl -n "${namespace}" get secrets "${secretName}" -o json | jq -e "select(.data[\"${clientIdKey}\"] != null)" &>/dev/null; then
-  echo "Unable to find k8s secret with key \"${secretKey}\" in \"${clientIdKey}\" for namespace \"${namespace}\""
-  exit 1
+  echo "Unable to find k8s secret with key \"${clientIdKey}\" in \"${secretName}\" for namespace \"${namespace}\""
+  [[ -z "${create_keys_in_secret}" ]] && echo "If this is not expected to exist then use the -k argument to create" && exit 1
```
mattops, line 113: I cant see where, when the `create_keys_in_secret` option is set, that the key/values are actually created?

rotate_clientID_cloudplatform_app.sh
petergphillips: yes, make sense, will move

rotate_clientID_cloudplatform_app.sh
petergphillips: the existing update secret will create the key/value if it doesn't exist already.  Therefore this flag is actually just to prevent it keeling over when they can't be found.  Have run it successfully to create the secret and also the key / value pairs.

mattops (approved):

PR #945, opened 2021-07-12
DT-2263: ✨ Filter clients

petergphillips:


src/main/resources/static/css/app.css
```diff
@@ -495,3 +495,43 @@
 .extra-wide .govuk-width-container {
   max-width: 1300px;
 }
+.horizontal-form * {
```
petergphillips, line 498: grabbed from prison staff hub

src/test/kotlin/uk/gov/justice/digital/hmpps/oauth2server/integration/ClientConfigSpecification.kt
```diff
@@ -347,30 +347,119 @@ class ClientConfigSpecification : AbstractAuthSpecification() {
...
+    find("table tbody tr")
+    this.role.text()
+    assertThat(rows).hasSizeGreaterThanOrEqualTo(rowsMin)
+    assertThat(rows).hasSizeLessThanOrEqualTo(rowsMax)
     assertThat(el("tr[data-qa='$client']").text()).startsWith(text.replaceIndent().replace("\n", ""))
     return this
   }
 
+  private val rows: FluentList<FluentWebElement>?
+    get() = find("table tbody tr")
```
petergphillips, line 461: no idea what is going on here, but having as a field on the page meant that it was caching old values.  Switching to a dynamic getter means that it calculates each time

SimonMitchellMOJ (approved):
LGTM

PR #947, opened 2021-07-13
DT-2263: ✨ Improve display of roles and grant types

petergphillips:


src/main/resources/static/css/govuk-frontend-3.6.0.min.css
```diff
@@ -1,3 +0,0 @@
-.govuk-link{font-family:GDS Transport,Arial,sans-serif;-webkit-font-smoothing:antialiased;-moz-osx-font-smoothing:grayscale}
```
petergphillips, line 1: old files no longer in use and hadn't realised that they were still around

JamieDraperUK (approved):
:+1:

PR #948, opened 2021-07-13
DT-2194: added new endpoint to get auth user details by userId and…

SimonMitchellMOJ:
…updated create user to return userId

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/maintain/AuthUserService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/maintain/AuthUserService.kt
@@ -7,6 +7,7 @@ import org.slf4j.LoggerFactory
 import org.springframework.beans.factory.annotation.Value
 import org.springframework.data.domain.Page
 import org.springframework.data.domain.Pageable
+import org.springframework.data.repository.findByIdOrNull
 import org.springframework.security.core.GrantedAuthority
 import org.springframework.security.crypto.password.PasswordEncoder
 import org.springframework.stereotype.Service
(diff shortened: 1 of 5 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/maintain/AuthUserService.kt
```diff
@@ -232,6 +235,8 @@ class AuthUserService(
   fun getAuthUserByUsername(username: String?): Optional<User> =
     userRepository.findByUsernameAndMasterIsTrue(StringUtils.upperCase(StringUtils.trim(username)))
 
+  fun getAuthUserByUserId(id: String): Optional<User> = userRepository.findById(UUID.fromString(id))
```
petergphillips, line 238: would be better to call `findByIdOrNull` here and then return `User?`

src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/resource/api/AuthUserController.kt
```diff
@@ -88,6 +88,33 @@ class AuthUserController(
...
+      ApiResponse(code = 401, message = "Unauthorized.", response = ErrorDetail::class),
+      ApiResponse(code = 404, message = "User not found.", response = ErrorDetail::class)
+    ]
+  )
+  fun getUserById(
+    @ApiParam(value = "The ID of the user.", required = true) @PathVariable userId: String,
+  ): ResponseEntity<Any?> {
+
+    val user = authUserService.getAuthUserByUserId(userId)
+    log.info("user found {}", user)
```
petergphillips, line 111: don't really need this log message

src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/resource/api/AuthUserController.kt
```diff
@@ -88,6 +88,33 @@ class AuthUserController(
...
+  @ApiResponses(
+    value = [
+      ApiResponse(code = 200, message = "OK", response = AuthUser::class),
+      ApiResponse(code = 401, message = "Unauthorized.", response = ErrorDetail::class),
+      ApiResponse(code = 404, message = "User not found.", response = ErrorDetail::class)
+    ]
+  )
+  fun getUserById(
+    @ApiParam(value = "The ID of the user.", required = true) @PathVariable userId: String,
+  ): ResponseEntity<Any?> {
```
petergphillips, line 108: we're trying to move away from doing this - would be better to return `AuthUser` instead and then throw a `UsernameNotFoundException` or similar which `AuthExceptionHandler` will then catch and return a `404` for us automatically.

petergphillips (dismissed):
LGTM

petergphillips (approved):
LGTM

PR #960, opened 2021-07-26
DT-1796: ✨ Convert alter user commands to stored procedures

petergphillips:


src/main/resources/db/nomis/V1_2__h2_procedures.sql
```diff
@@ -0,0 +1,33 @@
+CREATE ALIAS lock_user AS $$
+import java.sql.Connection;
+import java.sql.SQLException;
+@CODE
+void lockUser(Connection conn, String username) throws SQLException {
```
petergphillips, line 5: By writing the java code in a sql file we avoid the veracode issue :grin:.
Functions created here mimic the ones in OMS_UTILS.
Next steps will be to get the alter user oracle version to call the OMS_UTILS stored procedures and get both to use a common repository with the callable statements declared there instead.

src/main/resources/db/nomis/V1_2__h2_procedures.sql
```diff
@@ -0,0 +1,40 @@
...
+    statement.executeUpdate();
+}
+$$;
+
+CREATE ALIAS unlock_user AS $$
+import java.sql.Connection;
+import java.sql.SQLException;
+@CODE
+void unlockUser(Connection conn, String username) throws SQLException {
+    // no action required as h2 does not have concept of locking / unlocking users
```
mikehalmamoj, line 19: Which makes sense but then why implement `lockUser`?  If it's for testing why not just revert `lockUser` in this procedure?

mikehalmamoj (approved):

PR #963, opened 2021-07-27
DT-1796: ♻️ Improve h2 stored procedures after review

petergphillips:


src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/nomis/repository/StaffUserAccountRepository.kt
```diff
@@ -18,4 +19,16 @@ interface StaffUserAccountRepository : CrudRepository<NomisUserPersonDetails, St
     firstName: String,
     lastName: String,
   ): List<NomisUserPersonDetails>
+
+  @Modifying
```
petergphillips, line 23: need `Modifying` here since calling stored procedure which doesn't return a result set

src/main/resources/db/nomis/V1_2__h2_procedures.sql
```diff
@@ -30,11 +34,20 @@ void changePassword(Connection conn, String username, String password) throws SQ
...
-    lockStatement.setString(1, "OPEN");
-    lockStatement.setString(2, username);
-    lockStatement.executeUpdate();
+    final var statusSql = "SELECT account_status from dba_users WHERE username = ?";
+    final var statusStatement = conn.prepareStatement(statusSql);
+    statusStatement.setString(1, username);
+    final var resultSet = statusStatement.executeQuery();
+    if (resultSet.next()) {
+        final var accountStatus = resultSet.getString(1);
+        if (accountStatus.equals("EXPIRED")) {
```
petergphillips, line 43: mimic oracle functionality in that password reset removes the expiry

mikehalmamoj (approved):

PR #965, opened 2021-07-27
DT-1796: ✨ Convert oracle version to call stored procedures too

petergphillips:


src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/security/NomisOracleAlterUserService.kt
```diff
@@ -17,46 +16,28 @@ import javax.sql.DataSource
...
-    jdbcTemplate.update(String.format(UPDATE_STATUS, username))
-  }
-
-  private fun changePassword(username: String?, password: String?, template: String) {
     try {
-      jdbcTemplate.update(String.format(template, username, password))
-    } catch (e: DataAccessException) {
-      if (e.cause is SQLException) {
-        val sqlException = e.cause as SQLException?
-        if (sqlException!!.errorCode == 28007) {
```
petergphillips, line 54: exceptions changed as result of using jpa

Mjwillis (approved):
LGTM

PR #968, opened 2021-07-28
DT-2251 replace /login with /sign-in (and /logout with /sign-out)

mikehalmamoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/resource/LoginController.kt
```diff
@@ -41,6 +42,9 @@ class LoginController(
   }
 
   @GetMapping("/login")
+  fun oldLoginPage() = RedirectView("sign-in").apply { setPropagateQueryParams(true) }
```
petergphillips, line 45: this is nice :+1:

petergphillips (approved):
LGTM

PR #970, opened 2021-07-29
DT-2194: ♻️ Remove unused endpoints

petergphillips:
I have left two endpoints in use by other teams:
- GET /auth/api/authuser/{username}
- GET /auth/api/authuser/{username}/groups

found by running
```
requests
| where cloud_RoleName == 'hmpps-auth'
| where url contains "authuser/"
| where url !contains "authuser/id"
| where customDimensions.clientId <> ```manage-hmpps-auth-accounts-1```
| distinct name
```


amend-email.sh
```diff
@@ -1,72 +0,0 @@
-#!/bin/bash
```
petergphillips, line 1: created but hasn't been used recently so removing.  Would have to change the script to go get the user id of the user and then call the amend user email endpoint so not worth the effort.

src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/resource/api/AuthUserController.kt
```diff
@@ -262,14 +260,14 @@ class AuthUserController(
 
     if (user != null) {
       return ResponseEntity.status(HttpStatus.CONFLICT)
-        .body(ErrorDetailUsername("username.exists", "User $email already exists", "username", user.username))
+        .body(ErrorDetailUsername("username.exists", "User $email already exists", "userId", user.userId))
```
petergphillips, line 263: this is only used by the `user-load.sh` script, needs changing to user id so that we can add roles to the user by `userId` instead

src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/resource/api/AuthUserController.kt
```diff
@@ -591,7 +407,7 @@ class AuthUserController(
         response = ErrorDetail::class
       ), ApiResponse(
         code = 403,
-        message = "Unable to amend user, the user is not within one of your groups",
+        message = "Unable to amend user, the user is not within one of your groups or you don't have ROLE_MAINTAIN_OAUTH_USERS or ROLE_AUTH_GROUP_MANAGER roles",
```
petergphillips, line 410: improvement in documentation from previous PR

src/test/kotlin/uk/gov/justice/digital/hmpps/oauth2server/resource/api/AuthUserControllerIntTest.kt
```diff
@@ -87,18 +87,18 @@ class AuthUserControllerIntTest : IntegrationTest() {
   }
 
   @Nested
-  inner class EnableUser {
+  inner class EnableUserByUserId {
```
petergphillips, line 90: when `EnableUserByUserId` endpoint was added, tests were accidentally not copied over, so change to `userId` endpoint here now

src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/resource/api/AuthUserController.kt
```diff
@@ -652,11 +468,11 @@ class AuthUserController(
...
     val error_description: String,
 
-    @ApiModelProperty(required = false, value = "Field in error", example = "username", position = 3)
+    @ApiModelProperty(required = false, value = "Field in error", example = "userId", position = 3)
     val field: String? = null,
 
-    @ApiModelProperty(required = false, value = "username", example = "username", position = 4)
-    val username: String? = null
+    @ApiModelProperty(required = false, value = "userId", example = "userId", position = 4)
+    val userId: String? = null
```
mikehalmamoj, line 475: The class is still called `ErrorDetailUsername` - but I don't know the context so maybe that genuinely doesn't need switching to UserId?

mikehalmamoj (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/resource/api/AuthUserController.kt
petergphillips: agreed, will fix in next PR

PR #978, opened 2021-08-11
DT-2432: 🐛 No point checking delius on create

petergphillips:


src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/resource/api/AuthUserController.kt
```diff
@@ -256,7 +256,7 @@ class AuthUserController(
 
     val email = EmailHelper.format(createUser.email)
 
-    val user = email?.let { userService.findMasterUserPersonDetails(email).orElse(null) }
+    val user = email?.let { authUserService.getAuthUserByUsername(email).orElse(null) }
```
petergphillips, line 259: just go to the auth database rather than checking all 3

Mjwillis (approved):
LGTM

PR #980, opened 2021-08-13
Add Backstage Config

swilson-mojhub:
Context

We are building an automated single source of truth service catalogue proof of concept using backsatge.io.

Intent

This is a small PR to add the metadata for the HMPPS Auth so that backstage.io can pick up the data and add the HMPPS Auth into the service catalogue


mikehalmamoj (approved):

hmpps-catalog-info.yaml
```diff
@@ -0,0 +1,23 @@
...
+      title: Health endpoint
+      icon: LocalHospital
+    - url: https://github.com/ministryofjustice/nomis-api-terraform-azure/
+      title: Infra code
+      icon: Build
+spec:
+  type: service
+  lifecycle: live
+  owner: dps-tech-team
+
```
petergphillips, line 23: missing `definition: $text` in the `spec` section compared to https://github.com/ministryofjustice/prison-api/blob/main/hmpps-catalog-info.yaml

swilson-mojhub: It doesn't need it in this instance b/c it is a `component`, I know Auth has an API but I was going to add it as an API so it surfaced in the API section of Backstage.io then link the two together.

PR #981, opened 2021-08-13
DT-2447 New endpoint to retrieve auth user emails

mikehalmamoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/resource/api/UserController.kt
```diff
@@ -178,6 +178,7 @@ class UserController(private val userService: UserService) {
     }
     .orElseGet { notFoundResponse(username) }
 
+  @Deprecated("Please user /api/authuser/email instead")
```
mikehalmamoj, line 181: Currently only being used by `manage-hmpps-auth-accounts` so will be removed once the UI has switched over to the new endpoint.

petergphillips (approved):
LGTM

PR #985, opened 2021-08-19
DT-2448: removed eager loading of roles from user. …

SimonMitchellMOJ:
…Replaced with @Query within repository where needed


petergphillips (approved):
LGTM

PR #987, opened 2021-08-23
DT-2434: add role description to roles, admin_type table and linking …

SimonMitchellMOJ:
…tables for role to service and role to admin type


petergphillips (dismissed):
LGTM

petergphillips (approved):
LGTM

PR #991, opened 2021-08-25
DT-2475 Allow reset of password for user whose email address has been updated in delius

karenmillermoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/verify/ResetPasswordServiceImpl.kt
```diff
@@ -77,10 +77,14 @@ class ResetPasswordServiceImpl(
...
     val user = optionalUser.get()
     val templateAndParameters = getTemplateAndParameters(url, multipleMatchesAndCanBeReset, user)
-    sendEmail(user.username, templateAndParameters, user.email!!)
+    sendEmail(user.username, templateAndParameters, user.masterEmail())
     return Optional.ofNullable(templateAndParameters.resetLink)
   }
 
+  private fun User.masterEmail(): String =
+    if (source == AuthSource.delius) userService.findMasterUserPersonDetails(email!!).map { it.toUser().email!! }.orElse(email)
+    else email!!
```
karenmillermoj, line 86: There are lot of !!

src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/verify/ResetPasswordServiceImpl.kt
```diff
@@ -77,10 +77,14 @@ class ResetPasswordServiceImpl(
     }
     val user = optionalUser.get()
     val templateAndParameters = getTemplateAndParameters(url, multipleMatchesAndCanBeReset, user)
-    sendEmail(user.username, templateAndParameters, user.email!!)
+    sendEmail(user.username, templateAndParameters, user.masterEmail())
     return Optional.ofNullable(templateAndParameters.resetLink)
   }
 
+  private fun User.masterEmail(): String =
+    if (source == AuthSource.delius) userService.findMasterUserPersonDetails(email!!).map { it.toUser().email!! }.orElse(email)
```
petergphillips, line 85: not sure if this is quite right - this is looking them up in the source system based on their email address (which we think is incorrect).  Should be instead calling
```  
userService.getMasterUserPersonDetails(username, source)
```
to look them up in delius as we know the user and source system already.  Note that method is `private` at present, but should be made public so that we can then call it

src/test/kotlin/uk/gov/justice/digital/hmpps/oauth2server/verify/ResetPasswordServiceTest.kt
```diff
@@ -115,6 +115,30 @@ class ResetPasswordServiceTest {
       assertThat(optional).isEmpty
     }
 
+    @Test
+    fun requestResetPassword_emailAddressDifferentInDelius() {
+      val user = createSampleUser(username = "USER", email = "email", person = Person("Bob", "Smith"), enabled = true, source = delius)
+      val userFromDelius = createSampleUser(username = "USER", email = "emailFromDelius", person = Person("Bob", "Smith"), enabled = true, source = delius)
+      whenever(userRepository.findByEmail(any())).thenReturn(listOf(user, user))
+      whenever(userService.findEnabledOrNomisLockedUserPersonDetails(anyString())).thenReturn(user)
+      whenever(userService.findMasterUserPersonDetails("email")).thenReturn(Optional.of(userFromDelius))
```
petergphillips, line 124: `findMasterUserPersonDetails` takes a username, not an email address so won't return anything in this case

petergphillips (approved):
LGTM

PR #1004, opened 2021-09-06
DT-2492: view role details

SimonMitchellMOJ:


src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/auth/model/Authority.kt
```diff
@@ -29,6 +37,14 @@ class Authority(roleCode: String, roleName: String, roleDescription: String = ""
   @Column(name = "role_description")
   var roleDescription: String?
 
+  @OneToMany
+  @JoinTable(
+    name = "role_admin_type",
```
petergphillips, line 42: will need to be careful of this - don't want to create extra joins when doing the user download etc

src/main/resources/db/auth/V5_49__add_role_admin_types.sql
```diff
@@ -0,0 +1,11 @@
+ALTER TABLE admin_type
```
petergphillips, line 1: wondering if this would be better as an enum in auth?

src/main/resources/db/auth/V5_49__add_role_admin_types.sql
```diff
@@ -0,0 +1,11 @@
...
+    ADD admin_type_code VARCHAR(30)      NOT NULL
+        CONSTRAINT admin_type_code_uk UNIQUE;
+
+
+INSERT INTO admin_type (admin_type_id,admin_type_code,admin_type_name)
+VALUES (newid(),'LSA','Local System Administrator'),
```
petergphillips, line 7: Worth prefixing these two with DPS so that it is clearer

src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/auth/model/Authority.kt
```diff
@@ -5,14 +5,20 @@ import org.hibernate.annotations.GenericGenerator
...
 import javax.persistence.Table
 
 @Entity
 @Table(name = "ROLES")
-class Authority(roleCode: String, roleName: String, roleDescription: String = "") : GrantedAuthority {
+class Authority(
+  roleCode: String,
+  roleName: String,
+  roleDescription: String? = null,
+  adminType: List<String> = listOf(),
```
petergphillips, line 20: was expecting this to be a `List<AdminType>` rather than list of strings - one for next PR?

src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/auth/model/Authority.kt
```diff
@@ -29,6 +35,10 @@ class Authority(roleCode: String, roleName: String, roleDescription: String = ""
   @Column(name = "role_description")
   var roleDescription: String?
 
+  @Column(name = "admin_type")
+  @Convert(converter = StringListConverter::class)
```
petergphillips, line 39: would need a string -> list of enum converter here instead

petergphillips (approved):
LGTM

PR #1006, opened 2021-09-07
DCS-1180 ca_ro_dm_user added

GurnankCheema:
New user created for HDC testing. The user needs all of ca, ro and dm roles.


src/main/resources/db/dev/data/nomis/V9_1__user_data.sql
```diff
@@ -512,6 +518,9 @@ VALUES (-1, 'MDI', 'ITAG_USER'),
   (49, 'NWEB', 'HWPV_ALL_USER'),
   (50, 'NWEB', 'HWPV_SSCL_USER'),
   (50, 'NWEB', 'HWPV_ALL_USER');
+  (11, 'NWEB', 'CA_RO_DM_USER'),
+  (12, 'NWEB', 'CA_RO_DM_USER'),
+  (13, 'NWEB', 'CA_RO_DM_USER'),
```
andrewrlee, line 523: add `8` global search too?

src/main/resources/db/dev/data/auth/V900_0__oauth_sample_data.sql
```diff
@@ -140,6 +140,7 @@ INSERT INTO users (user_id, username, email, verified, source)
         ('C3B15C4B-ADF5-493B-9424-DBCC65E7BFED', 'UOF_COORDINATOR_USER', 'uof_coordinator@digital.justice.gov.uk', 1, 'nomis'),
         ('C3B15C4B-ADF5-493B-9424-DBCC65E7BFEE', 'LICENCE_READONLY_TEST', 'licence_readonly_test@digital.justice.gov.uk', 1, 'nomis'),
         ('C3B15C4B-ADF5-493B-9424-DBCC65E8BFEF', 'RO_USER_READONLY_TEST', 'ro_user_readonly_test@digital.justice.gov.uk', 1, 'nomis'),
+        ('C3B15C4B-ADF5-493B-9424-DBCC65E8BFEF', 'CA_RO_DM_USER', 'ca_ro_dm_user_test@digital.justice.gov.uk', 1, 'nomis'),
```
bell-pa, line 143: I think this creates an auth user. Do you really want this?  It has the same username as the NOMIS user and also the same user_id as RO_USER_READONLY_TEST above it.  That can't be right.  Probably breaks a unique constraint on the users table too.

src/main/resources/db/dev/data/nomis/V9_1__user_data.sql
```diff
@@ -512,6 +518,9 @@ VALUES (-1, 'MDI', 'ITAG_USER'),
...
```
bell-pa, line 524: You might also need GLOBAL_SEARCH when acting in the RO role.  Depends upon what you want to do with the user.

src/main/resources/db/dev/data/nomis/V9_1__user_data.sql
GurnankCheema: ok

src/main/resources/db/dev/data/auth/V900_0__oauth_sample_data.sql
GurnankCheema: hmm. Not totally sure if I need an auth user. So I'll remove as suggested.

bell-pa (approved):
LGTM

src/main/resources/db/dev/data/nomis/V9_1__user_data.sql
petergphillips: we're trying to move away from global search role, prefer `VIEW_PRISONER_DATA`

andrewrlee: this is to match an existing type of user who will already have global search

PR #1007, opened 2021-09-07
DT-2493 Allow user to amend a role name

karenmillermoj:


petergphillips (dismissed):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/resource/api/RolesController.kt
```diff
@@ -151,3 +182,10 @@ data class RoleAdminType(
     AdminType.valueOf(a.uppercase(getDefault())).description
   )
 }
+
+@ApiModel(description = "Role Name")
+data class RoleAmendment(
+  @ApiModelProperty(required = true, value = "Role Name", example = "Central admin")
+  @field:NotBlank(message = "Role name must be supplied")
```
petergphillips, line 189: should really validate the min and max length here too (between 4 and 100?)

karenmillermoj: validation is in the ui side but can add here too

SimonMitchellMOJ (approved):
LGTM

PR #1009, opened 2021-09-07
DT-2492: update adminType so that it can be serialized …

SimonMitchellMOJ:
…and create EnumListConverter for getting adminTypes from database


src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/auth/model/Authority.kt
```diff
@@ -65,3 +67,16 @@ class Authority(
...
   }
 }
+
+@Converter
+class EnumListConverter : AttributeConverter<List<AdminType>, String> {
+  override fun convertToDatabaseColumn(stringList: List<AdminType>): String =
+    stringList.joinToString(",") { it.adminTypeCode.trim() }
+
+  override fun convertToEntityAttribute(string: String): List<AdminType> {
+    return string.split(",").map {
```
petergphillips, line 77: how does this work for existing data in the database where the data is `null`?

petergphillips (approved):
LGTM

PR #1014, opened 2021-09-09
NN-3641 adding a redirect url to the test data

richardpopple:


Mjwillis (commented):
Should this be sign-in now not login?

richardpopple:
> Should this be sign-in now not login?

When I spin this all up locally then the prisonstaffhub sends the call back url of  http://localhost:3002/login/callback. 

Also when I navigate to:
https://digital-dev.prison.service.justice.gov.uk
I get bounced to:
https://sign-in-dev.hmpps.service.justice.gov.uk/auth/sign-in?redirect_uri=https://digital-dev.prison.service.justice.gov.uk/login/callback
So I think this is consistent. 
But I might be out of the loop on this one and maybe there should have been a change from login to sign-in?

PR #1016, opened 2021-09-09
PIC-1655 Pre-sentence Service

theDustRoom:
hmpps-template-typescript has changed and now uses `/sign-in` instead of `/login` so running the app locally does not work.

🗃️ Include pre-sentence-service

Signed-off-by: theDustRoom <paul.massey@digital.justice.gov.uk>


SimonMitchellMOJ (approved):
LGTM

PR #1022, opened 2021-09-13
DT-2509 Additional test for regex validation

karenmillermoj:


karenmillermoj:
The formatting has gone a bit awry due to moving create role into a nested class

petergphillips (approved):
LGTM

PR #1024, opened 2021-09-16
DT-2568 Allow role admin types to be updated

karenmillermoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/resource/api/RolesController.kt
```diff
@@ -260,13 +285,20 @@ data class RoleNameAmendment(
...
   @field:Pattern(regexp = "^[0-9A-Za-z- ,.()'&\r\n]*\$")
-  val roleDescription: String?,
+  val roleDescription: String?
+)
+
+@ApiModel(description = "Role Administration Types")
+data class RoleAdminTypeAmendment(
+  @ApiModelProperty(required = true, value = "Role Administration Types", example = "[\"DPS_ADM\"]")
+  @field:NotEmpty(message = "Admin type cannot be empty")
+  val adminType: MutableSet<AdminType>
```
mikehalmamoj, line 303: Is there any reason this needs to be a mutable set?

mikehalmamoj: Aha1 just spotted why :smile: - might need to add DPS_ADM type too. Rather than making this a mutable set it would be better to take a copy and add the extra type in the new instance.

mikehalmamoj (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/resource/api/RolesController.kt
karenmillermoj: Yes - you're right - will do another PR to update that

PR #1027, opened 2021-09-21
DT-2571: 🐛 Suppress false positive for spring security jwt

petergphillips:


auth-suppressions.xml
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<suppressions xmlns="https://jeremylong.github.io/DependencyCheck/dependency-suppression.1.3.xsd">
+  <suppress>
+    <notes><![CDATA[
+    file name: spring-security-jwt-1.1.1.RELEASE.jar
```
petergphillips, line 6: vulnerability is in spring security instead.  Matcher is
```
cpe:2.3:a:pivotal_software:spring_security:*:*:*:*:*:*:*:* versions up to (excluding) 5.2.9
```
which appears to match this unnecessarily

andymarke (approved):
LGTM

PR #1030, opened 2021-09-23
MMP-258 - Add manage my prison local client and nomis role and user

garethmdavies:
Manage my Prison service client plus user and role for running locally against dockerised hmpps-auth


brightonsbox (approved):
LGTM

PR #1059, opened 2021-10-29
DT-2803: ✨ Switch lock and unlock over to using nomis user api

petergphillips:


src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/nomis/service/NomisUserApiService.kt
```diff
@@ -9,10 +9,24 @@ class NomisUserApiService(
   @Qualifier("nomisWebClient") private val webClient: WebClient,
 ) {
   fun changePassword(username: String, password: String) {
-    webClient.put().uri("/{username}/change-password", username)
+    webClient.put().uri("/users/{username}/change-password", username)
```
petergphillips, line 12: turns out I missed `/users` prefix from the change password endpoint

SimonMitchellMOJ (approved):
LGTM

PR #1063, opened 2021-11-03
DT-2813 Handle 400 errors on change password

andymarke:


src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/nomis/service/NomisUserApiService.kt
```diff
@@ -30,3 +48,13 @@ class NomisUserApiService(
...
+
+fun <T> errorWhenBadRequest(
+  exception: WebClientResponseException,
+  errorMapper: (content: String) -> AuthenticationException
+): Mono<T> =
+  if (exception.rawStatusCode == BAD_REQUEST.value()) Mono.error(errorMapper(exception.responseBodyAsString)) else Mono.error(
+    exception
+  )
+
+data class PasswordChangeError(val errorCode: Int? = 0)
```
andymarke, line 60: oddly I needed to give `errorCode` a default value else Jackson claimed there was no default constructor, not sure what is missing from Auth since I'm sure I've not had to do that before

src/test/kotlin/uk/gov/justice/digital/hmpps/oauth2server/nomis/service/NomisUserApiServiceTest.kt
```diff
@@ -25,16 +35,123 @@ class NomisUserApiServiceTest : IntegrationTest() {
 
   @BeforeEach
   fun setUp() {
-    nomisService = NomisUserApiService(webClient)
+    nomisService = NomisUserApiService(webClient, ObjectMapper().configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false))
```
petergphillips, line 38: would be better to use
```
    @Autowired
    private ObjectMapper objectMapper;
```
instead so you get the one created by spring rather than creating one manually

andymarke: 100%, I'll change that now

petergphillips (approved):
LGTM

PR #1075, opened 2021-11-12
NN-3654 adding the managing adjudications reviewer role

richardpopple:


petergphillips (changes requested):
Sorry, roles are no longer added via this process for production.  You can add development data here but new roles should go via #dps_appsupport

PR #1082, opened 2021-11-19
DT-2860: add service details to client details page

SimonMitchellMOJ:


src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/resource/ClientsController.kt
```diff
@@ -47,12 +50,14 @@ class ClientsController(
   fun showEditForm(@RequestParam(value = "client", required = false) baseClientId: String?): ModelAndView {
     return if (baseClientId != null) {
       val (clientDetails, clients) = clientService.loadClientWithCopies(baseClientId)
-      val clientDeployment =
-        clientService.loadClientDeploymentDetails(baseClientId) ?: ClientDeployment(baseClientId = baseClientId)
+      val clientDeployment = clientService.loadClientDeploymentDetails(baseClientId) ?: ClientDeployment(baseClientId = baseClientId)
+      val serviceDetails = authServicesService.loadServiceDetails(baseClientId) ?: Service(code = baseClientId, name = "", description = "", url = "")
```
petergphillips, line 54: I think the parameter to this method is incorrect - shouldn't it be `clientId` instead?

src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/resource/ClientsController.kt
```diff
@@ -47,12 +50,14 @@ class ClientsController(
...
       ModelAndView("ui/form", "clientDetails", AuthClientDetails(clientDetails as BaseClientDetails))
         .addObject("clients", clients)
         .addObject("deployment", clientDeployment)
         .addObject("baseClientId", ClientService.baseClientId(baseClientId))
+        .addObject("service", serviceDetails)
+        .addObject("baseClientId", ClientService.baseClientId(baseClientId))
```
petergphillips, line 60: two `baseClientId` here?  Also why do we need to translate `baseClientId`, unless it is just `clientId` instead?

src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/service/AuthServicesService.kt
```diff
@@ -21,6 +21,11 @@ class AuthServicesService(private val oauthServiceRepository: OauthServiceReposi
   fun getService(code: String): Service = oauthServiceRepository.findById(code)
     .orElseThrow { EntityNotFoundException("Entity $code not found") }
 
+  fun loadServiceDetails(clientId: String): Service? {
+    val searchClientId = ClientService.baseClientId(clientId)
```
petergphillips, line 25: would be better to translate this earlier on

petergphillips (approved):
LGTM

PR #1083, opened 2021-11-23
DT-2896: updated rotation script so that it doesn't try to restart deployment…

SimonMitchellMOJ:
…when we have used -d ignore missing deployment param


petergphillips (approved):
LGTM

PR #1096, opened 2021-11-30
DT-2923 Add generic exception handler

andymarke:
* Required since runtime errors are not being logged
* Also  fixed NPE that had no test


petergphillips (approved):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/config/AuthExceptionHandler.kt
```diff
@@ -173,6 +176,29 @@ class AuthExceptionHandler {
...
+  fun rethrowAccessException(e: HttpMessageConversionException) {
+    throw e
+  }
+
+  @ExceptionHandler(ServletRequestBindingException::class)
+  fun rethrowAccessException(e: ServletRequestBindingException) {
+    throw e
+  }
+
+  @ExceptionHandler(Exception::class)
```
andymarke, line 194: Once this is added, the above is necessary so Spring caches and converts the correct response code

petergphillips (approved):
LGTM

PR #1097, opened 2021-11-30
DT-2923 Allow to break request into chunks

andymarke:
There seems to be a 2 minutes timeout in Azure gateway, so this work around allows us to break the requests down by 200 users at a time.

30 Pages at a time should be ok, so will need to run 7 or so requests (and inspect logs to see how many pages are allowed)


petergphillips (approved):
LGTM

PR #1098, opened 2021-12-01
DT-2923 Switch Auth to use stage that has better staff data

andymarke:
community-api stage uses preprod for tokens so this should also be ok


Mjwillis (approved):
lgtm

PR #1102, opened 2021-12-03
DT-2947: call NOMIS User Roles API from changeEmailAndRequestVerifica…

neilmendum:
https://dsdmoj.atlassian.net/browse/DT-2947


src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/nomis/service/NomisUserApiService.kt
```diff
@@ -110,6 +111,18 @@ class NomisUserApiService(
...
   }
 
+  fun findUserByUsername(username: String): Optional<NomisApiUserPersonDetails> {
+    if (!nomisEnabled) {
+      log.debug("Nomis integration disabled, returning empty for {}", username)
+      return Optional.empty()
+    }
+    val userDetails = webClient.get().uri("/users/{username}", username)
+      .retrieve()
+      .bodyToMono(object : ParameterizedTypeReference<NomisApiUserDetails>() {})
```
petergphillips, line 121: you don't need to use `ParameterizedTypeReference` - that is only for `List` and other collections iirc

src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/nomis/service/NomisUserApiService.kt
```diff
@@ -110,6 +111,18 @@ class NomisUserApiService(
...
+      .block()!!
+    return Optional.of(mapUserDetailsToNomisUser(userDetails))
```
petergphillips, line 123: don't you need to handle the 404 here?

src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/nomis/service/NomisUserApiService.kt
neilmendum: Good spot - addressed

src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/nomis/service/NomisUserApiService.kt
```diff
@@ -110,6 +111,18 @@ class NomisUserApiService(
       .block()!!
   }
 
+  fun findUserByUsername(username: String): Optional<NomisApiUserPersonDetails> {
```
petergphillips, line 114: we're trying to replace all `Optional` with `null` in kotlin, so might be better to use `NomisApiUserPersonDetails?` here instead

neilmendum: Gotcha. Addressed

src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/security/NomisUserService.kt
```diff
@@ -53,17 +53,17 @@ abstract class NomisUserService(
...
+      return verifyEmailService.changeEmailAndRequestVerification(
+        username = it.username,
+        emailInput = emailInput,
+        firstName = it.firstName,
+        fullname = it.name,
+        url = url,
+        emailType = PRIMARY
+      )
+    }
+    throw UsernameNotFoundException("Account for username $username not found")
```
petergphillips, line 66: you could use an elvis operator here and then convert the function to an expresson body

src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/nomis/service/NomisUserApiService.kt
neilmendum: Addressed

src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/security/NomisUserService.kt
neilmendum: Addressed

src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/security/NomisUserService.kt
```diff
@@ -49,19 +52,24 @@ abstract class NomisUserService(
...
+    getNomisUserByUsernameFromNomisUserApiService(username)?.let {
+      return verifyEmailService.changeEmailAndRequestVerification(
+        username = it.username,
+        emailInput = emailInput,
+        firstName = it.firstName,
+        fullname = it.name,
+        url = url,
+        emailType = PRIMARY
+      )
+    } ?: run {
```
petergphillips, line 70: don't think you need `run {}` here, should just be able do to the `throw`?

src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/nomis/service/NomisUserApiService.kt
```diff
@@ -110,6 +110,34 @@ class NomisUserApiService(
...
+        log.debug("User not found in NOMIS due to {}", it.message)
+        Mono.empty()
+      }
+      .onErrorResume(WebClientResponseException::class.java) {
+        log.warn(
+          "Unable to retrieve details from NOMIS for user {} due to {}",
+          username,
+          (it as WebClientResponseException).statusCode
+        )
+        Mono.empty()
```
petergphillips, line 133: This doesn't look quite right - we're saying that if we're unable to call the endpoint then we will just `warn` on the exception and carry on.  Was expecting the exception still to be thrown as we can't call nomis for that user.  Not so important now, but will be later on when we add the authenticate etc.

neilmendum: Good point. Changed that by following what was done with `DeliusAuthenticationServiceException` so there's consistency between the two web clients

src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/security/NomisUserService.kt
neilmendum: Good spot, addressed

petergphillips (approved):
LGTM

PR #1108, opened 2021-12-07
DT-2978: 🐛 Cope with calling api for email

petergphillips:


src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/verify/ResetPasswordServiceImpl.kt
```diff
@@ -115,11 +115,15 @@ class ResetPasswordServiceImpl(
     val user = userPersonDetails.toUser()
     return if (!passwordAllowedToBeReset(user, userPersonDetails)) {
       Optional.empty()
-    } else userService.getEmailAddressFromNomis(user.username).map { e: String? ->
-      user.email = e
-      user.verified = true
-      userRepository.save(user)
-      user
+    } else {
+      val email = if (userPersonDetails is NomisApiUserPersonDetails) Optional.ofNullable(userPersonDetails.email)
```
petergphillips, line 119: we can come in with both type of nomis user here at present, so have to cope with them

src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/verify/VerifyEmailService.kt
```diff
@@ -287,17 +287,6 @@ class VerifyEmailService(
              and s.username = :username
       """
 
-    private const val EXISTING_EMAIL_FOR_USERNAMES_SQL =
```
petergphillips, line 290: not needed as rest of code removed

Mjwillis (approved):
LGTM

PR #1109, opened 2021-12-07
DT-2981: call NOMIS user roles API from findEnabledOrNomisLockedUserPersonDetails

neilmendum:
https://dsdmoj.atlassian.net/browse/DT-2981


src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/security/UserService.kt
```diff
@@ -50,7 +50,11 @@ class UserService(
 
   fun findEnabledOrNomisLockedUserPersonDetails(username: String): UserPersonDetails? =
     authUserService.getAuthUserByUsername(username).filter { it.isEnabled }.map { UserPersonDetails::class.java.cast(it) }
-      .or { nomisUserService.getNomisUserByUsername(username).filter { it.isEnabled || it.accountDetail.status == AccountStatus.LOCKED }.map { UserPersonDetails::class.java.cast(it) } }
+      .or {
+        Optional.ofNullable(nomisUserService.getNomisUserByUsernameFromNomisUserApiService(username))
```
neilmendum, line 54: Because all of the other userServices still return `Optional` I decided to keep the implementation using `Optional.or` and `Optional.filter` the same here, hence using `ofNullable(nomisUserService...`

petergphillips (approved):
LGTM

PR #1110, opened 2021-12-09
DT-2981: use webClient with hmpps-auth-nomis-api client creds instead…

neilmendum:
https://dsdmoj.atlassian.net/browse/DT-2981


petergphillips (approved):
LGTM

PR #1113, opened 2021-12-10
DT-2978: 🐛 Fix email parameter query param

petergphillips:


src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/nomis/service/NomisUserApiService.kt
```diff
@@ -85,8 +85,8 @@ class NomisUserApiService(
     }
     val userDetails = webClient.post().uri {
       it.path("/users/user")
-        .queryParam("email", emailAddress)
-        .build()
+        .queryParam("email", "{emailAddress}")
+        .build(emailAddress)
```
petergphillips, line 89: doing it this way means that the query parameter is encoded, otherwise it is not.

andymarke (approved):
LGTM

PR #1114, opened 2021-12-10
DT-2979: ✨ Prevent users from accessing service if no permissions

petergphillips:


src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/resource/UserSelectorAndMfaAuthorizationEndpoint.kt
```diff
@@ -93,7 +93,7 @@ class UserSelectorAndMfaAuthorizationEndpoint(
     val mfaPassed = user?.passedMfa == true
 
     val replacedAuthentication = if ((clientNeedsMfa && mfaPassed) || !clientNeedsMfa) {
-      if (!account.isNullOrBlank() && account != "true" && authentication != null) {
```
petergphillips, line 96: no longer required as only came through in the no roles scenario

src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/security/UserContextApprovalHandler.kt
```diff
@@ -89,13 +89,13 @@ class UserContextApprovalHandler(
         roles = serviceRoles
       )
       userApprovalRequest["users"] = users
-      userApprovalRequest["service"] = service?.name
+      userApprovalRequest["service"] = service?.name ?: "this service"
```
petergphillips, line 92: one for future PRs I think - the only scenario where the service is `null` here is if it is an azure account that hasn't been mapped.  Not clear what should happen in that case...

andymarke (approved):
LGTM

PR #1116, opened 2021-12-13
DT-2925: add call to authenticate endpoint in NOMIS User Roles API

neilmendum:


petergphillips (approved):
LGTM

PR #1117, opened 2021-12-14
DT-2925: map user flags from NOMIS User Roles API UserDetail

neilmendum:

Map changes from https://github.com/ministryofjustice/nomis-user-roles-api/pull/73


src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/nomis/model/NomisApiUserDetails.kt
```diff
@@ -9,7 +9,9 @@ data class NomisApiUserDetails @JsonCreator constructor(
   @JsonProperty("lastName") val surname: String,
   @JsonProperty("firstName") val firstName: String,
   @JsonProperty("primaryEmail") val email: String?,
-  @JsonProperty("active") val enabled: Boolean,
   @JsonProperty("dpsRoleCodes") val roles: List<String>,
   @JsonProperty("accountStatus") val accountStatus: AccountStatus,
+  @JsonProperty("isAccountNonLocked") val isAccountNonLocked: Boolean,
+  @JsonProperty("isCredentialsNonExpired") val isCredentialsNonExpired: Boolean,
+  @JsonProperty("isEnabled") val isEnabled: Boolean
```
petergphillips, line 16: this looks a bit odd - isn't the property `enabled` instead

neilmendum: That's a good point. Have updated this PR and will raise another PR for `nomis-user-roles-api`

petergphillips (approved):
LGTM

PR #1120, opened 2021-12-15
DT-2925: call NOMIS user roles API from NomisUserDetailsService and NomisAuthenticationProvider

neilmendum:


src/test/kotlin/uk/gov/justice/digital/hmpps/oauth2server/integration/ChangeExpiredPasswordSpecification.kt
```diff
@@ -67,20 +62,21 @@ class ChangeExpiredPasswordSpecification : AbstractNomisAndDeliusAuthSpecificati
...
       .inputAndConfirmNewPassword("helloworld2", "helloworld2")
-    homePage
-      .isAtPage()
-      .logOut()
-    loginPage
-      .isAtPage()
-      .loginAs("EXPIRED_TEST2_USER", "helloworld2")
-    homePage
-      .isAtPage()
-      .assertNameDisplayedCorrectly("C. Password2")
```
neilmendum, line 78: Because the responsibility for updating the password and authentication is now handled by NOMIS User Roles API we are only validating the http calls made to it. I did think about using [stateful WireMock](http://wiremock.org/docs/stateful-behaviour/) responses in this test to validate the re-login after the password had been updated, but didn't think it added much value

src/test/kotlin/uk/gov/justice/digital/hmpps/oauth2server/security/NomisAuthenticationProviderTest.kt
```diff
@@ -0,0 +1,90 @@
...
+import org.mockito.kotlin.mock
+import org.mockito.kotlin.verify
+import org.mockito.kotlin.whenever
+import org.springframework.security.authentication.LockedException
+import org.springframework.security.authentication.UsernamePasswordAuthenticationToken
+import uk.gov.justice.digital.hmpps.oauth2server.nomis.model.NomisUserPersonDetailsHelper
+import uk.gov.justice.digital.hmpps.oauth2server.nomis.service.NomisUserApiService
+import uk.gov.justice.digital.hmpps.oauth2server.service.MfaClientNetworkService
+
+class NomisAuthenticationProviderTest {
```
neilmendum, line 18: Replaces `NomisAuthenticationProviderIntTest` because it was a `SpringBootTest` and didn't use the `NomisExtension` so we couldn't mock the http responses. Converted most of the relevant tests to _Mockito style_ and checked it tested the same things as `DeliusAuthenticationProviderTest`

src/test/kotlin/uk/gov/justice/digital/hmpps/oauth2server/integration/LoginSpecification.kt
```diff
@@ -246,12 +246,6 @@ class LoginSpecification : AbstractNomisAndDeliusAuthSpecification() {
       .checkError("Enter a valid username and password. You will be locked out if you enter the wrong details 3 times.")
   }
 
-  @Test
-  fun `Attempt login with expired user wrong password`() {
```
neilmendum, line 250: Dropped this one because if the password is wrong then `authenticate` fails and so we don't grab the user details from NOMIS User Roles API

src/test/kotlin/uk/gov/justice/digital/hmpps/oauth2server/integration/LoginSpecification.kt
```diff
@@ -182,7 +182,7 @@ class LoginSpecification : AbstractNomisAndDeliusAuthSpecification() {
 
   @Test
   fun `Log in fails with valid credentials same user name in Auth and Nomis but both accounts disabled`() {
-    goTo(loginPage).loginError("NOMIS_LOCKED_AUTH_DISABLED", "password123456")
+    goTo(loginPage).loginError("LOCKED_USER", "password123456")
```
petergphillips, line 185: not sure about this one - by changing it to `LOCKED_USER` doesn't it mean that it won't be checking auth any more?

src/test/kotlin/uk/gov/justice/digital/hmpps/oauth2server/integration/VerifyEmailSpecification.kt
```diff
@@ -77,6 +77,8 @@ class VerifyEmailSpecification : AbstractAuthSpecification() {
 
   @Test
   fun `A user can verify an email that exists in pnomis`() {
+    // Given RO_USER does not have a record in V900_0__oauth_sample_data.sql so the email needs to be verified
+    // and find_users_ro_user.json does not contain an email so the NomisApiUserPersonDetails.toUser sets verified to false
```
petergphillips, line 81: the test name doesn't correspond to this test then - it suggests that an email that exists in pnomis needs verifying

src/test/kotlin/uk/gov/justice/digital/hmpps/oauth2server/integration/LoginSpecification.kt
neilmendum: Good point. Reverted that

src/test/kotlin/uk/gov/justice/digital/hmpps/oauth2server/integration/VerifyEmailSpecification.kt
neilmendum: Good point. Renamed it to `A user needs to be verified if they don not have an email configured in pnomis`

petergphillips (dismissed):
LGTM

petergphillips (approved):
LGTM

PR #1121, opened 2021-12-16
SLM-53 Fix client credential roles

mikehalmamoj:
Turns out if there is a space after the comma then subsequent roles are not included!


SimonMitchellMOJ (approved):
LGTM

PR #1122, opened 2021-12-16
DT-2878 Indicate that password reset affect legacy ID providers of NOMIS and nDelius

Mjwillis:
- DT-2878 Display legacy ID system on password reset


lbennett-moj (approved):
LGTM

PR #1125, opened 2021-12-17
DT-2925: prefix NOMIS roles with ROLE_

neilmendum:


src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/nomis/model/NomisApiUserPersonDetails.kt
```diff
@@ -41,7 +44,19 @@ data class NomisApiUserPersonDetails(
 
   override fun eraseCredentials() {}
 
-  override fun getAuthorities(): Collection<GrantedAuthority?> = roles.plus(SimpleGrantedAuthority("ROLE_PRISON"))
+  override fun getAuthorities(): Collection<GrantedAuthority?> {
+    val rolesRoleWithPrefix = roles.stream().map {
```
neilmendum, line 48: Assumed that unlike [NomisUserPersonDetails](https://github.com/ministryofjustice/hmpps-auth/blob/main/src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/nomis/model/NomisUserPersonDetails.kt#L93) we wouldn't need to filter out `NWEB` because the roles returned here are not a collection of `UserCaseloadRole`

petergphillips: don't need to `stream` - code was originally copied from the java

src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/nomis/model/NomisApiUserPersonDetails.kt
```diff
@@ -41,7 +44,19 @@ data class NomisApiUserPersonDetails(
...
+      SimpleGrantedAuthority(
+        "ROLE_" + StringUtils.upperCase(
```
petergphillips, line 50: pretty ugly kotlin here, would be better to use string interpolation and kotlin `uppercase` function

src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/nomis/model/NomisApiUserPersonDetails.kt
```diff
@@ -41,7 +44,19 @@ data class NomisApiUserPersonDetails(
...
+          it?.authority?.replace(
+            '-',
+            '_'
+          )
+        )
+      )
+    }
+    return Stream.concat(rolesRoleWithPrefix, setOf(SimpleGrantedAuthority("ROLE_PRISON")).stream()).collect(Collectors.toSet())
```
petergphillips, line 58: since you're not doing a `stream`, the `plus` will be fine

petergphillips (dismissed):
LGTM

petergphillips (approved):
LGTM

PR #1126, opened 2021-12-20
DT-2998: call NOMIS user roles API from findMasterUserPersonDetails and getMasterUserPersonDetails

neilmendum:
https://dsdmoj.atlassian.net/browse/DT-2998

And map field added with https://github.com/ministryofjustice/nomis-user-roles-api/pull/80


src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/nomis/model/NomisApiUserDetails.kt
```diff
@@ -8,10 +8,12 @@ data class NomisApiUserDetails @JsonCreator constructor(
...
   @JsonProperty("firstName") val firstName: String,
+  @JsonProperty("activeCaseloadId") val activeCaseloadId: String?,
   @JsonProperty("primaryEmail") val email: String?,
   @JsonProperty("dpsRoleCodes") val roles: List<String>,
   @JsonProperty("accountStatus") val accountStatus: AccountStatus,
   @JsonProperty("accountNonLocked") val accountNonLocked: Boolean,
   @JsonProperty("credentialsNonExpired") val credentialsNonExpired: Boolean,
-  @JsonProperty("enabled") val enabled: Boolean
+  @JsonProperty("enabled") val enabled: Boolean,
+  @JsonProperty("admin") val admin: Boolean
```
petergphillips, line 18: in kotlin land we prefer the last parameter to also have a `,` on the end, that way when new parameters are added it doesn't look like the previous line was changed too

petergphillips (approved):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/nomis/model/NomisApiUserDetails.kt
neilmendum: That's good to know. I'll try and remember it next time, am still in Java land but slowly getting there 😄

PR #1128, opened 2022-01-04
DT-2928: drop displaying a list of emails from email verification page

neilmendum:
https://dsdmoj.atlassian.net/browse/DT-2928

With #1120 we introduced logic for `NomisApiUserPersonDetails.verified = !email.isNullOrEmpty()` because if the user has an email associated with their NOMIS account then it is now provided by NOMIS User Roles API. Previously auth used `NomisUserPersonDetails` which mapped to `STAFF_USER_ACCOUNTS` and we had to do a separate lookup to find the email addresses for that user

I raised https://github.com/ministryofjustice/nomis-user-roles-api/pull/83 then after discussing with @petergphillips we decided to change the verify email screen - to remove list of emails being populated with `verifyEmailService.getExistingEmailAddressesForUsername`. The user will either:
- Already be verified because they have an email address associated with their NOMIS account
- Not be verified because they do not have an email address associated with their NOMIS account, and need to provide one using the verify email form


src/main/resources/templates/verifyEmail.html
```diff
@@ -32,49 +32,14 @@ <h1 class="govuk-heading-l">Verify your email address</h1>
...
         <p class="govuk-body">
           We've found an existing email address that hasn't been verified yet. Please amend the email address if
           necessary and we'll send another verification email.
         </p>
       </div>
 
-      <div th:class="'govuk-form-group' + (${error} ? ' govuk-form-group--error' : '')"
-           th:if="${#lists.isEmpty(candidates)}">
+      <div th:class="'govuk-form-group' + (${error} ? ' govuk-form-group--error' : '')">
         <input type="hidden" id="candidate" value="other" name="candidate"/>
```
petergphillips, line 43: note we pass `candidate` set to `other` here - can be removed.  The `PostMapping` in `VerifyEmailController` can then have the `candidate` field removed too (note it was previously coming in as `other` not blank)

petergphillips: also referenced in https://github.com/ministryofjustice/hmpps-auth/blob/2f006cd9abcc9247e36e59165b4bbb2dad4d6162/src/main/resources/templates/changeEmail.html#L43 and also https://github.com/ministryofjustice/hmpps-auth/blob/2f006cd9abcc9247e36e59165b4bbb2dad4d6162/src/main/resources/templates/account/changeBackupEmail.html#L37 which post to that same post action

neilmendum: Thanks for flagging this. I've renamed `candidate` to `changeType` so it's hopefully clearer what the form field is used for

src/test/kotlin/uk/gov/justice/digital/hmpps/oauth2server/integration/VerifyEmailSpecification.kt
```diff
@@ -174,12 +185,6 @@ open class VerifyEmailPage :
     verifyEmailButton.click()
   }
 
-  fun selectExistingEmailAs(email: String) {
```
neilmendum, line 177: No longer required as we don't provide `candidates` to the view

src/test/kotlin/uk/gov/justice/digital/hmpps/oauth2server/integration/VerifyEmailSpecification.kt
```diff
@@ -92,6 +92,17 @@ class VerifyEmailSpecification : AbstractAuthSpecification() {
     verifyEmailConfirmPage.isAt()
   }
 
+  @Test
+  fun `A user is shown an error if they do not provide an email`() {
+    // Given RO_USER2 does not have a record in V900_0__oauth_sample_data.sql so the email needs to be verified
+    // and find_users_ro_user_2.json does not contain an email so NomisApiUserPersonDetails.toUser sets verified to false
+    goTo(loginPage).loginAsWithUnverifiedEmail("RO_USER2")
```
neilmendum, line 99: The test _A user needs to be verified if they do not have an email configured in pnomis_ has already verified the email for `RO_USER` hence introducing `RO_USER2`

src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/resource/VerifyEmailController.kt
```diff
@@ -88,8 +84,8 @@ class VerifyEmailController(
 
   @PostMapping("/verify-email")
   fun verifyEmail(
-    @RequestParam(required = false) candidate: String,
-    @RequestParam email: String?,
+    @RequestParam changeType: String,
+    @RequestParam email: String,
```
petergphillips, line 88: why can't this be `null` anymore?

src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/resource/VerifyEmailController.kt
```diff
@@ -107,21 +103,19 @@ class VerifyEmailController(
       }
     }
 
-    // candidate will either be an email address from the selection or 'other' meaning free text
-    if (StringUtils.isEmpty(candidate)) {
-      return verifyEmailRequest(principal, request, response, "noselection")
+    if (StringUtils.isEmpty(email)) {
+      return verifyEmailRequest(principal, request, response, "blank")
```
petergphillips, line 107: why the new special break out here for `blank`?  Should already be catered for in the validation in `changeEmailAndRequestVerification`

neilmendum: Good point, addressed

src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/resource/VerifyEmailController.kt
petergphillips: If it comes through as `null` then it won't even get to the validation, however I'm guessing since the functional tests still succeed then I guess it is coming through as empty instead?

petergphillips (approved):
LGTM

PR #1132, opened 2022-01-07
DT-3019: tidy up getNomisUserByUsername

neilmendum:
https://dsdmoj.atlassian.net/browse/DT-3019


src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/security/UserService.kt
```diff
@@ -54,7 +54,7 @@ class UserService(
   fun findEnabledOrNomisLockedUserPersonDetails(username: String): UserPersonDetails? =
     authUserService.getAuthUserByUsername(username).filter { it.isEnabled }.map { UserPersonDetails::class.java.cast(it) }
       .or {
-        Optional.ofNullable(nomisUserService.getNomisUserByUsernameFromNomisUserApiService(username))
+        Optional.ofNullable(nomisUserService.getNomisUserByUsername(username))
```
matt-crowson-moj, line 57: Minor comment. May be good to look at removing or limit use Optionals going forwards since Kotlin provides a way of achieving this with null-able types. Of course avoiding nulls altogether would be great but it is never that simple and Spring repos can and do return Optional types for some of the default methods also.

matt-crowson-moj (approved):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/security/UserService.kt
neilmendum: Good point @matt-crowson-moj thanks for raising. For new methods we're trying to avoid Optionals but haven't yet gone back to refactor the existing usages

PR #1131, opened 2022-01-06
DT-3019: remove StaffUserAccountRepository

neilmendum:
https://dsdmoj.atlassian.net/browse/DT-3019

PR #1134, opened 2022-01-10
DT-3029: remove NOMIS JDBC datasource

neilmendum:
https://dsdmoj.atlassian.net/browse/DT-3029


src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/security/AuthUserDetailsService.kt
```diff
@@ -16,7 +16,11 @@ import javax.persistence.EntityManager
 import javax.persistence.PersistenceContext
 
 @Service("authUserDetailsService")
-@Transactional(transactionManager = "authTransactionManager", readOnly = true)
+@Transactional(
+  transactionManager = "authTransactionManager",
+  readOnly = true,
+  noRollbackFor = [UsernameNotFoundException::class]
```
neilmendum, line 22: Added this because it was breaking `OauthIntTest` tests see [analysis on Slack](https://mojdt.slack.com/archives/C028V9TS295/p1641999944021700). While I didn't get to the root cause of why removing `NomisDbConfig` caused the transaction to rollback, I think adding `noRollbackFor` is a fair solution because
- This transaction boundary is already marked as `readOnly` which means Hibernate doesn't commit it
- In `loadUserByUsername` we detach the `authEntityManager` to avoid committing any changes to user details

Therefore introducing it will not affect the flow of transactions this class is part of

helm_deploy/values-t3.yaml
```diff
@@ -17,7 +17,7 @@ mfa_allowlist:
   global_protect_vpn: "35.176.93.186/32"
 
 env:
-  SPRING_PROFILES_ACTIVE: "oracle,nomis,delius,token-verification,azure-oidc"
+  SPRING_PROFILES_ACTIVE: "nomis,delius,token-verification,azure-oidc"
```
Mjwillis, line 20: is the nomis profile needed

Mjwillis (approved):
LGTM - see comment

helm_deploy/values-t3.yaml
neilmendum: that's a good point and I was thinking about it too - in https://github.com/ministryofjustice/hmpps-auth/pull/1131#discussion_r779699052 I decided to keep it to be consistent with the `delius` profile. what do you think?

PR #1138, opened 2022-01-13
DT-3019: map active flag from NOMIS User Roles API

neilmendum:
https://dsdmoj.atlassian.net/browse/DT-3019


src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/verify/ResetPasswordServiceImpl.kt
```diff
@@ -183,7 +183,7 @@ class ResetPasswordServiceImpl(
     }
     val nomisApiUser = userPersonDetails as NomisUserPersonDetails
     val status = nomisApiUser.accountStatus
-    return nomisApiUser.isEnabled && (!status.isLocked || status.isUserLocked || user.locked)
+    return nomisApiUser.active && (!status.isLocked || status.isUserLocked || user.locked)
```
neilmendum, line 186: With https://github.com/ministryofjustice/hmpps-auth/pull/1108/ we changed the logic for `passwordAllowedToBeReset `
- from
  - `staffUserAccount.staff.isActive`
- to
  - `nomisApiUser.isEnabled`

I had a look and for a locked account 
- `staff.isActive = true `
- `user.isEnabled = false`

Which means the method started returning false. Reintroducing `active` to fix this so we can reset a locked NOMIS account

SimonMitchellMOJ (approved):
LGTM

PR #1140, opened 2022-01-14
DT-2950 When changing email address (for Nomis User) only save in Nom…

karenmillermoj:
…is when email has been verified


Mjwillis (approved):
LGTM

PR #1141, opened 2022-01-14
DT-3050: use spring prefix for properties

neilmendum:


helm_deploy/hmpps-auth/templates/_envs.tpl
```diff
@@ -147,21 +147,6 @@ env:
       secretKeyRef:
         key: JWT_SIGNING_KEY_PAIR
         name: {{ template "app.name" . }}
-  - name: SPRING_DATASOURCE_PASSWORD
-    valueFrom:
-      secretKeyRef:
-        key: SPRING_DATASOURCE_PASSWORD
```
neilmendum, line 153: We can delete `SPRING_DATASOURCE_*` secrets from K8

src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/config/AuthDbConfig.kt
```diff
@@ -19,14 +19,14 @@ import javax.sql.DataSource
 @EnableJpaRepositories(
   basePackages = ["uk.gov.justice.digital.hmpps.oauth2server.auth.repository"]
 )
-@ConfigurationProperties(prefix = "auth")
+@ConfigurationProperties(prefix = "spring")
```
Mjwillis, line 22: is that need still `spring`

neilmendum: I tried dropping that then `DataSourceBuilder` couldn't resolve property names, I did a bit of tweaking but couldn't figure it out so thought I'd leave it

PR #1142, opened 2022-01-14
DT-3050: nest jpa properties correctly

neilmendum:


src/main/resources/application.yml
```diff
@@ -29,19 +29,22 @@ spring:
...
-      hbm2ddl:
-        auto: none
   session:
     store-type=jdbc
   data:
     web:
       pageable:
         max-page-size: 20000
 
+  jpa:
```
neilmendum, line 39: With #1134 and #1141 it looks like I combined `jpa.hibernate` and `jpa.properties.hibernate` which I think caused issues, see jira

SimonMitchellMOJ (approved):
LGTM

PR #1143, opened 2022-01-17
DT-3050: show sql queries to help with debugging

neilmendum:


src/main/resources/application.yml
```diff
@@ -43,7 +43,7 @@ spring:
     properties:
       hibernate:
         dialect: org.hibernate.dialect.H2Dialect
-        show_sql: false
+        show_sql: true
```
Mjwillis, line 46: You don't want to do this.  It will log logs in prod.  Instead env field SPRING_JPA_PROPERTIES_HIBERNATE_SHOW_SQL=true

Mjwillis (changes requested):

PR #1144, opened 2022-01-17
DT-2498 Add ability to get all users and emails using AuthSource

karenmillermoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/resource/api/AuthUserController.kt
```diff
@@ -453,6 +454,32 @@ class AuthUserController(
...
+    produces = "application/json"
+  )
+  @ApiResponses(
+    value = [
+      ApiResponse(code = 200, message = "OK", response = AuthUser::class),
+      ApiResponse(code = 401, message = "Unauthorized.", response = ErrorDetail::class),
+      ApiResponse(code = 404, message = "User not found.", response = ErrorDetail::class)
+    ]
+  )
+  @PreAuthorize("hasAnyRole('ROLE_MAINTAIN_ACCESS_ROLES', 'ROLE_MAINTAIN_ACCESS_ROLES_ADMIN')")
```
karenmillermoj, line 472: I think this role should be ROLE_MAINTAIN_OAUTH_USERS (or similar) - will update

src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/resource/api/AuthUserController.kt
```diff
@@ -453,6 +454,32 @@ class AuthUserController(
...
+  fun getAllUserEmails(
+    @ApiParam(value = "A single auth source to search [nomis|delius|auth|azuread]. Defaults to auth if omitted.")
```
neilmendum, line 474: The endpoints is this class look they handle requests for auth users only, perhaps this one should live in `UserController` as that's for users of all types

karenmillermoj: It's funny because I must've moved it between user<->authuser about 10 times!
Will update...cheers

neilmendum: Haha! np

neilmendum (approved):
LGTM

PR #1150, opened 2022-01-26
HAAR-14 Run Auth with a postgres auth db

karenmillermoj:


src/main/resources/application-postgres-config.yml
```diff
@@ -0,0 +1,5 @@
+spring:
+  datasource:
+    url: 'jdbc:h2:mem:authdb;MODE=PostgreSQL'
+  flyway:
+    locations: db/auth_{vendor}
```
neilmendum, line 5: We did this to override `application.yml` which sets locations to `db/auth,db/auth_{vendor}`

`db/auth` contains inserts which aren't compatible with Postgres so we decided to introduce `db/auth_postgresql` because we wanted to leave the default configuration as SQLServer, or H2 (with SQLServer dialect) when running locally

It also has the history of table changes & inserts which aren't relevant for Postgres because we will start fresh, construct every table with `V1_0__postgres_schema.sql`, then copy over all table data with a database export

src/main/resources/db/auth_postgresql/V1_0__postgres_schema.sql
```diff
@@ -0,0 +1,271 @@
+    create table groups
```
neilmendum, line 1: Nabbed from #1044

neilmendum: Updated with changes applied to `db/auth` since that PR was created

neilmendum (approved):

PR #1152, opened 2022-01-28
HAAR-86: update configuration to deploy stage to CloudPlatform

neilmendum:
https://dsdmoj.atlassian.net/browse/HAAR-86

PR #1153, opened 2022-02-01
HAAR-86: configure datasource and jpa dialect in each env

neilmendum:


helm_deploy/hmpps-auth/templates/_envs.tpl
```diff
@@ -47,23 +47,14 @@ env:
...
   - name: SPRING_DATASOURCE_USERNAME
-    valueFrom:
-      secretKeyRef:
-        key: AUTH_DATASOURCE_USERNAME
-        name: {{ template "app.name" . }}
+    value: {{ .Values.env.SPRING_DATASOURCE_USERNAME | quote }}
 
   - name: SPRING_JPA_HIBERNATE_DIALECT
-    value: "org.hibernate.dialect.SQLServer2012Dialect"
+    value: {{ .Values.env.SPRING_JPA_HIBERNATE_DIALECT | quote }}
```
neilmendum, line 57: With #1152 I assumed it was possible to override `_envs.tpl` from a particular env config but that's not correct hence now configuring each env

Mjwillis (approved):
LGTM

PR #1154, opened 2022-02-01
HAAR-86: do not nest namespace_secrets under env

neilmendum:


karenmillermoj (approved):
lgtm

PR #1155, opened 2022-02-01
Revert "HAAR-86: configure datasource and jpa dialect in each env (#1…

neilmendum:
…153)"

This reverts commit a8cd6c7e


karenmillermoj (approved):
lgtm

PR #1156, opened 2022-02-01
HAAR-86: add separate helm chart directory so can use generic-service chart when deploying to cloud platform

neilmendum:


helm_deploy_cp/hmpps-auth/values.yaml
```diff
@@ -0,0 +1,85 @@
...
+
+  namespace_secrets:
+    hmpps-auth:
+      APPINSIGHTS_INSTRUMENTATIONKEY: "APPINSIGHTS_INSTRUMENTATIONKEY"
+      APPLICATION_GOOGLE_TAG_ID: "APPLICATION_GOOGLE_TAG_ID"
+      APPLICATION_NOTIFY_KEY: "APPLICATION_NOTIFY_KEY"
+      DELIUS_CLIENT_CLIENT_ID: "DELIUS_CLIENT_CLIENT_ID"
+      DELIUS_CLIENT_CLIENT_SECRET: "DELIUS_CLIENT_CLIENT_SECRET"
+      NOMIS_CLIENT_CLIENT_ID: "NOMIS_CLIENT_CLIENT_ID"
+      NOMIS_CLIENT_CLIENT_SECRET: "NOMIS_CLIENT_CLIENT_SECRET"
```
neilmendum, line 67: Copied from #1053 which includes 71b7638f4aef86278413d969a3f25786d0981b4f. I added `NOMIS_CLIENT` properties which were introduced afterwards

.circleci/config.yml
```diff
@@ -118,6 +118,7 @@ workflows:
       - hmpps/deploy_env:
           name: deploy_stage
           env: "stage"
+          helm_dir: "helm_deploy_cp"
```
neilmendum, line 121: This is `helm_deploy` by default. Providing it so we can configure the `generic-service` chart without affecting the existing deployments

helm_deploy_cp/hmpps-auth/values.yaml
```diff
@@ -0,0 +1,85 @@
...
+      JWT_KEYSTORE_PASSWORD: "JWT_KEYSTORE_PASSWORD"
+      JWT_SIGNING_KEY_PAIR: "JWT_SIGNING_KEY_PAIR"
+      TOKENVERIFICATION_CLIENT_CLIENT_ID: "TOKENVERIFICATION_CLIENT_CLIENT_ID"
+      TOKENVERIFICATION_CLIENT_CLIENT_SECRET: "TOKENVERIFICATION_CLIENT_CLIENT_SECRET"
+      AUTH_AZUREOIDC_CLIENT_ID: "AUTH_AZUREOIDC_CLIENT_ID"
+      AUTH_AZUREOIDC_CLIENT_SECRET: "AUTH_AZUREOIDC_CLIENT_SECRET"
+      AUTH_AZUREOIDC_TENANT_ID: "AUTH_AZUREOIDC_TENANT_ID"
+
+    dps-rds-instance-output:
```
karenmillermoj, line 76: No biggie but we try to use hmpps instead of dps - obv changing this would mean a cloud platform change too, but just something to note.

helm_deploy_cp/hmpps-auth/values.yaml
```diff
@@ -0,0 +1,85 @@
+---
+# Values here are the same across all environments
+generic-service:
+  nameOverride: hmpps-auth
+  replicaCount: 3
+
```
karenmillermoj, line 6: looks like we can pull this out - so it will use generic template value default of 4

helm_deploy_cp/hmpps-auth/values.yaml
```diff
@@ -0,0 +1,92 @@
...
+  readinessProbe:
+    httpGet:
+      path: /auth/health/readiness
+    periodSeconds: 20
+    initialDelaySeconds: 60
+    timeoutSeconds: 30
+    failureThreshold: 15
+
+  env:
+    JAVA_OPTS: "-Xmx512m"
```
neilmendum, line 35: We don't set `Xmx` in our existing charts so want to be a bit careful here otherwise could impact performance. 512m is [typical for other APIs](https://github.com/search?l=YAML&p=2&q=org%3Aministryofjustice+JAVA_OPTS&type=Code) but will check in Prometheus to see how much memory this app generally uses

karenmillermoj (approved):
lgtm

PR #1159, opened 2022-02-03
HAAR-86: correct env variable name in UUIDCustomType

neilmendum:


karenmillermoj (approved):
lgtm

PR #1161, opened 2022-02-04
HAAR-130: Spring profile to seed local Postgres instance

neilmendum:
https://dsdmoj.atlassian.net/browse/HAAR-130


src/main/resources/application-auth-seed-postgres.yml
```diff
@@ -0,0 +1,3 @@
+spring:
+  flyway:
+    locations: db/auth_postgresql,db/dev/data/auth-postgresql
```
neilmendum, line 3: This is different from `auth-seed` because that profile is compatible with both SQLServer and H2. Post migration we'll be able to neaten this all up with [HAAR-85](https://dsdmoj.atlassian.net/browse/HAAR-85)

karenmillermoj (approved):
lgtm

SimonMitchellMOJ (approved):
LGTM

PR #1181, opened 2022-03-09
HAAR-168:️ add allowed ips table for clients

SimonMitchellMOJ:
added field to client maintenance page to add/edit allowed ips


Mjwillis (approved):
LGTM

PR #1182, opened 2022-03-10
HAAR-233 Set limits to auth for memory 768 Mb

Mjwillis:


helm_deploy_cp/hmpps-auth/values.yaml
```diff
@@ -75,8 +75,16 @@ generic-service:
     timeoutSeconds: 30
     failureThreshold: 15
 
+    resources:
+      requests:
+        cpu: 100m
+        memory: 640Mi
+      limits:
+        cpu: 2000m
+        memory: 2048Mi
```
petergphillips, line 84: don't think this needs to be this high

petergphillips (approved):
LGTM

PR #1187, opened 2022-03-21
HAAR-170:️ added Validation to createAccessToken that a current client request is coming from a valid/allowed IP

SimonMitchellMOJ:


Mjwillis (approved):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/security/AuthIpSecurity.kt
```diff
@@ -21,4 +22,14 @@ class AuthIpSecurity(@Value("\${application.authentication.ui.allowlist}") priva
     }
     return matchIp
   }
+
+  @Throws(AllowedIpException::class)
+  fun validateClientIpAllowed(remoteIp: String?, clientAllowList: List<String>): Boolean {
```
karenmillermoj, line 27: I don't think you need to return a boolean anymore.
Because we are throwing an exception if it fails, it means that anyone calling this method will just need to ensure the exception isn't thrown.  i.e. we would never get a return matchIp value of false.

src/test/kotlin/uk/gov/justice/digital/hmpps/oauth2server/security/AuthIpSecurityTest.kt
```diff
@@ -1,43 +1,80 @@
...
-  }
+  @Nested
+  inner class IpCheck {
+    @Test
+    fun testStandardV4IP() {
+      val request = MockHttpServletRequest()
+      request.remoteAddr = "127.0.0.1"
+      val testClass = AuthIpSecurity(setOf("0.0.0.0/0"))
+      val check = testClass.check(request)
+      assertThat(check).isTrue
```
karenmillermoj, line 19: With reference to the previous comment, we should now just be able to call the check method.  (If it fails the exception should be thrown)

PR #1199, opened 2022-04-01
HAAR-301 resilience imporovements 

Mjwillis:
- Reduce timeouts for Auth DB to 1 sec /0.5 se validation
- Increase connection pool to 20
- reduce timeouts to Nomis user roles api to 2 seconds
- Increase pods to 8 from 4 (temp)


karenmillermoj (approved):
lgtm
I have some preprod and prod yamls ready to push back for the new cp so will ensure they are up to date too

PR #1201, opened 2022-04-04
HAAR-297 Rename user_contact value column to details and reset first postgres db

karenmillermoj:


src/main/resources/db/auth_postgresql/V1_0__create_tables.sql
```diff
@@ -268,4 +268,13 @@ create table email_domain
...
\ No newline at end of file
+);
+
+create table oauth_client_allowed_ips
+(
+    base_client_id varchar(64) not null
+        constraint oauth_client_allowed_ips_pk
+            primary key,
+    allowed_ips varchar(1000)
+
```
karenmillermoj, line 279: I have been able to change this file as I have cleared out the current dbs that are set to use postgres.  The new db will have the new column name for user_contact (details) and also this table will be created here (rather than needing a separate file)

SimonMitchellMOJ (approved):
LGTM

PR #1217, opened 2022-04-19
HAAR-328: update mod sec for preprod/prod

SimonMitchellMOJ:


helm_deploy_cp/values-devcp.yaml
```diff
@@ -6,49 +6,6 @@ generic-service:
   ingress:
     host: sign-in-dev.hmpps.service.justice.gov.uk
     modsecurity_audit_enabled: true
-    modsecurity_snippet: |
```
petergphillips, line 9: what's this change?

SimonMitchellMOJ: I only added the `SecAction` to dev so had to copy all the modsecurity_snippet in this now puts it all into the values.yaml file

PR #1227, opened 2022-04-22
SDI-85: ✨ Only do MFA for check my diary users with special role

petergphillips:


src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/service/MfaClientService.kt
```diff
@@ -11,9 +12,20 @@ open class MfaClientService(
...
 ) {
 
-  open fun clientNeedsMfa(clientId: String?): Boolean {
-    val client = clientDetailsService.loadClientByClientId(clientId)
+  open fun clientNeedsMfa(request: AuthorizationRequest?): Boolean {
+    val client = clientDetailsService.loadClientByClientId(request?.clientId)
+    val baseClientId = ClientService.baseClientId(client.clientId)
+
+    if (baseClientId == "my-diary" &&
+      request?.authorities?.none { it.authority == "ROLE_CMD_MIGRATED_MFA" } == true
```
petergphillips, line 20: check my diary users need `ROLE_CMD_MIGRATED_MFA` to go down the mfa route

Mjwillis (approved):
Nice

PR #1235, opened 2022-04-26
HAAR-361 Allow temporary access to old auth

karenmillermoj:


karenmillermoj:
No longer required.  The two prisons having issues (HAAR-361) now have access to the new cp deployment/postgres db

karenmillermoj:
No longer required.  The two prisons having issues (HAAR-361) now have access to the new cp deployment/postgres db

PR #1238, opened 2022-04-27
SDI-113: ✨ Check mfa remember me during 2fa process

petergphillips:


src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/service/MfaClientService.kt
```diff
@@ -25,6 +29,12 @@ open class MfaClientService(
...
 
     val mfa = client.additionalInformation["mfa"] as? String?
-    return (mfa == untrusted.name && mfaClientNetworkService.outsideApprovedNetwork()) || mfa == all.name
+    val requireMfa = (mfa == untrusted.name && mfaClientNetworkService.outsideApprovedNetwork()) || mfa == all.name
+    if (!requireMfa) return false
+
+    // now check if they are allowed to use a remember me token (from the mfa_remember_me cookie) for that client
+    // and if it is still valid
+    val rememberMe = client.additionalInformation["mfaRememberMe"] as? Boolean? == true
+    return !(rememberMe && tokenService.checkToken(TokenType.MFA_RMBR, MfaRememberMeContext.token, user?.username))
```
petergphillips, line 38: we have to mark the client as not needing MFA since we only want the remember me to be applicable where the client has that configuration set.  This also means that the user hasn't actually MFA, it is more that it wasn't appropriate for this client this time round.  This means that if the user tries to access other services that have MFA enabled then they will be prompted to go through MFA for those services.

src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/utils/MfaRememberMeCookieFilter.kt
```diff
@@ -0,0 +1,36 @@
...
+
+@Component
+@Order(2)
+class MfaRememberMeCookieFilter(
+  private val mfaRememberMeCookieHelper: MfaRememberMeCookieHelper,
+) : OncePerRequestFilter() {
+
+  override fun doFilterInternal(request: HttpServletRequest, response: HttpServletResponse, filterChain: FilterChain) {
+    val cookieValue = mfaRememberMeCookieHelper.readValueFromCookie(request).orElse(null)
+    MfaRememberMeContext.token = cookieValue
```
petergphillips, line 19: it is a bit naff having to store the cookie in a threadlocal, but we can't access the cookie value otherwise from within the depths of the spring oauth security configuration

SimonMitchellMOJ (approved):
LGTM

PR #1249, opened 2022-05-03
SDI-86 user self serve account creation (& remove user authentication)

steverendell:
Develop Script to migrate check-my-diary auth data as well as report it


petergphillips (approved):
LGTM

PR #1267, opened 2022-05-10
SDI-86: ✨ Expose which mfa options are set

petergphillips:


src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/resource/api/UserController.kt
```diff
@@ -100,6 +100,29 @@ class UserController(private val userService: UserService) {
...
 
+  @GetMapping("/api/user/me/mfa")
+  @ApiOperation(
+    value = "MFA options configured for current user",
+    nickname = "myMfa",
+    consumes = "application/json",
+    produces = "application/json",
+  )
+  fun myMfa(@ApiIgnore principal: Principal): MfaOptions = userService
+    .getOrCreateUser(principal.name)
```
steverendell, line 111: create? can it insert in some circumstances?

petergphillips: well it more means that if the user doesn't exist in the auth database then it is added, rather than the user being created from scratch.  So if the user exists in nomis or delius for example but has never signed in before then they will be added automatically by this process

PR #1284, opened 2022-05-19
VB-805 Migration roles added to BPV client

philip-milne:
Additional client roles for development and testing book a prison visit migration


Mjwillis (approved):
Good luck!

PR #1285, opened 2022-05-20
HAAR-465 Timeout to NOMIS user roles too strict - up from 2 -> 8 seconds

Mjwillis:
Signed-off-by: mjwillis <michael.willis@digital.justice.gov.uk>


SimonMitchellMOJ (approved):
LGTM

PR #1287, opened 2022-05-20
VB-806 Add UI roles to client in HMPPS Auth db dev scripts and client…

AledWEvans:
This is to allow auth to be used locally using docker instance


Mjwillis (approved):
LGTM

PR #1296, opened 2022-05-27
HAAR-476: API to request NOMIS email sync for username

brightonsbox:


src/test/kotlin/uk/gov/justice/digital/hmpps/oauth2server/resource/api/PrisonUserControllerTest.kt
```diff
@@ -11,11 +11,13 @@ import org.mockito.kotlin.whenever
 import uk.gov.justice.digital.hmpps.oauth2server.security.NomisUserService
 import uk.gov.justice.digital.hmpps.oauth2server.security.PrisonUserDto
 import uk.gov.justice.digital.hmpps.oauth2server.security.UserService
+import uk.gov.justice.digital.hmpps.oauth2server.verify.VerifyEmailService
 
 class PrisonUserControllerTest {
```
hyderkhanmoj, line 16: Could probably add a unit test here for the new controller method

hyderkhanmoj (approved):
LGTM

src/test/kotlin/uk/gov/justice/digital/hmpps/oauth2server/resource/api/PrisonUserControllerTest.kt
brightonsbox: There's no logic in the controller so I've just tested in the integration test version

PR #1297, opened 2022-05-27
HAAR-491: introduce new Spring Security core OAuth library

AlTurner-MOJ:


Mjwillis (approved):
So assume libraries not used but don't clash

AlTurner-MOJ:
@Mjwillis yes, this is first step just to confirm that addition of libraries in itself does not break anything

PR #1306, opened 2022-06-01
HAAR-501: added in groovy 4.0.2 dependency due to owasp failure with …

SimonMitchellMOJ:
…thymeleaf-layout-dialect:3.1.0 which uses groovy 4.0.0 that contains CVE-2020-17521


brightonsbox (approved):
LGTM

PR #1307, opened 2022-06-01
HAAR-492: Should not show username for an azure user as this is a UUID

ravmoj:
Should not show username for an azure user as this is a UUID


Mjwillis (approved):
Lgtm

Mjwillis (approved):

PR #1313, opened 2022-06-08
HAAR-499: Correcting link to add service details once client rotated

brightonsbox:
Also removed geb config file since no longer used AFAIK


Mjwillis (approved):
LGTM!

PR #1317, opened 2022-06-13
HAAR-443: update client maintenance screen to use client valid for x …

SimonMitchellMOJ:
…days instead of end date


karenmillermoj (dismissed):
lgtm

Posss some kotlin .let for some of the null check code and is there a min/max on the validDays ?

PR #1319, opened 2022-06-20
HAAR-548: allow client expire check box uncheck also removes valid days

SimonMitchellMOJ:


Mjwillis (approved):
Approved, wasn't quite sure what the logic was doing in the comment however

PR #1331, opened 2022-07-08
HAAR-613: Validate email domain

ravmoj:
Added new endpoint to validate email domains


src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/service/EmailDomainService.kt
```diff
@@ -59,6 +59,16 @@ class EmailDomainService(
...
 
+  fun isValidEmailDomain(emailDomain: String): Boolean {
+    val domainNameInternal = if (emailDomain.startsWith(PERCENT)) emailDomain else PERCENT + emailDomain
+    val existingDomain = emailDomainRepository.findByName(domainNameInternal)
+
+    if (existingDomain != null) {
+      return true
+    }
+    return false
+  }
```
karenmillermoj, line 70: can probably just tidy this up to return !=null

src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/service/EmailDomainService.kt
```diff
@@ -59,6 +59,16 @@ class EmailDomainService(
...
+
   @Throws(EmailDomainNotFoundException::class)
   fun removeDomain(id: String) {
```
karenmillermoj, line 73: Not your change, but wondering why this doesn't already pass in a UUID
Can the controller be changed to pass in UUID instead of a string?

karenmillermoj (dismissed):
Looks good - minor tidy up but otherwise ok
Wonder if it's worth looking at changing the remove endpoint to only allow a UUID (obv this is not your work...but will mean endpoint validation is more specific to what we are passing in)

src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/service/EmailDomainService.kt
ravmoj: Updated

src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/service/EmailDomainService.kt
ravmoj: Updated

PR #1335, opened 2022-07-11
HAAR-628 Route Call from HMPPS Auth to token verification service internal network.

Mjwillis:
Signed-off-by: mjwillis <michael.willis@digital.justice.gov.uk>

PR #1336, opened 2022-07-12
Revert "HAAR-628 Route Call from HMPPS Auth to token verification service internal network."

Mjwillis:
Reverts ministryofjustice/hmpps-auth#1335


andrewrlee (approved):
lgtm

PR #1340, opened 2022-07-14
Revert "Revert "Revert "HAAR-628 Route Call from HMPPS Auth to token verification service internal network."""

SimonMitchellMOJ:
Reverts ministryofjustice/hmpps-auth#1338


Mjwillis (approved):
LGTM

PR #1364, opened 2022-08-03
HAAR-674:  password reset suspended user

ravmoj:
password reset suspended user


src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/nomis/model/StaffStatus.kt
```diff
@@ -0,0 +1,21 @@
+package uk.gov.justice.digital.hmpps.oauth2server.nomis.model
+
+enum class StaffStatus(
```
Mjwillis, line 3: Nervous about this.  Not sure its safe to use an ENUM when this data can change

src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/verify/ResetPasswordServiceImpl.kt
```diff
@@ -183,7 +184,7 @@ class ResetPasswordServiceImpl(
       AuthSource.nomis -> {
         val nomisApiUser = userPersonDetails as NomisUserPersonDetails
         val status = nomisApiUser.accountStatus
-        nomisApiUser.active && (!status.isLocked || status.isUserLocked || user.locked)
+        (nomisApiUser.staffStatus != StaffStatus.SUS) && (!status.isLocked || status.isUserLocked || user.locked)
```
Mjwillis, line 187: Are you sure this is right?  I don't think so, only should check if not ACTIVE

src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/nomis/model/StaffStatus.kt
Mjwillis: Really think this should be a string

Mjwillis (changes requested):
Would question use of an ENUM as this is reference data in the db

Mjwillis (approved):
LGTM

PR #1371, opened 2022-08-25
HAAR-729: 📝 updated swagger titles

SimonMitchellMOJ:


src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/resource/api/TokenController.kt
```diff
@@ -19,7 +19,7 @@ import javax.validation.Valid
 
 @RestController
 @Validated
-@Tag(name = "token-controller", description = "Token Controller")
+@Tag(name = "/api/new-token", description = "Token Controller")
```
karenmillermoj, line 22: I remember adding these in when I did the initial change - to match the original documentation ...I think you're right though - reads much better

karenmillermoj (approved):
lgtm

PR #1380, opened 2022-09-12
HAAR-726 Remove general Exception catch from exception handler to allow UI to fail gracefully through error page

karenmillermoj:


andymarke (approved):
LGTM

PR #1393, opened 2022-10-03
HAAR-752: split out nomis specification in int tests

SimonMitchellMOJ:
Separate the auth and nomis specifications into separate classes so that we can test nomis connection being down but auth still working - currently they are together which means we cannot


src/test/kotlin/uk/gov/justice/digital/hmpps/oauth2server/integration/DeliusDownLoginSpecification.kt
```diff
@@ -4,7 +4,7 @@ import org.assertj.core.api.Assertions
 import org.fluentlenium.core.annotation.Page
 import org.junit.jupiter.api.Test
 
-class DeliusDownLoginSpecification : AbstractAuthSpecification() {
+class DeliusDownLoginSpecification : AbstractNomisAuthSpecification() {
```
AlTurner-MOJ, line 7: Should this extend AbstractDeliusAuthSpecification?

SimonMitchellMOJ: no, we want Delius to be unavailable. but Nomis to be available. having AbstractDeliusAuthSpecification would make Delius available.

AlTurner-MOJ (approved):
lgtm

PR #1394, opened 2022-10-05
HAAR-819: Deprecate group '/api/groups' and '/api/groups/child/{group…

ravmoj:
Deprecate group '/api/groups' and '/api/groups/child/{group}' endpoints


karenmillermoj (approved):
lgtm

PR #1397, opened 2022-10-10
HAAR-826: Deprecate create child group ('/api/groups/child')

ravmoj:
Deprecate create child group ('/api/groups/child')


karenmillermoj (approved):
lgtm

PR #1410, opened 2022-11-04
HAAR-973: update to springboot plugin 4.5.7

SimonMitchellMOJ:
Commented out `AccountDetailsSpecification - azure account details` test as familyName and givenName are not present in the test but are required - testing in dev environment to see if it is test data that is incorrect since spring security 5.7.5


petergphillips (approved):
LGTM

PR #1414, opened 2022-11-16
HAAR-973: Added in scopes to test data token required by spring security 5.7.5 update

SimonMitchellMOJ:


src/test/kotlin/uk/gov/justice/digital/hmpps/oauth2server/integration/AccountDetailsSpecification.kt
```diff
@@ -66,20 +67,13 @@ class AccountDetailsSpecification : AbstractNomisAndDeliusAuthSpecification() {
     goTo(accountDetailsPage).checkNomisDetails()
   }
 
-  // commented out to test if spring security 5.7.5 changes,
```
petergphillips, line 69: don't think you need this commented out section anymore?

petergphillips (approved):
LGTM

PR #1420, opened 2022-11-21
HAAR-950: Added child_group to alter script

ravmoj:
HAAR-950: Added child_group to alter script and constraint on user_group, user_role and group_assignable_role


karenmillermoj (approved):
lgtm

PR #1432, opened 2022-12-09
HAAR-843: Deprecated Verified email address for users 

ravmoj:
Deprecated Verified email address for users


AlTurner-MOJ (dismissed):
lgtm

AlTurner-MOJ (commented):
As discussed, this end point is obsolete in Auth and therefore can be removed

PR #1433, opened 2022-12-12
HAAR-877 deprecate all /api/authuser/id/{userId}/roles endpoints as n…

karenmillermoj:
…ow moved to manageUsersApi


ravmoj (approved):
lgtm

PR #1439, opened 2022-12-21
HAAR-884 Add Find Azure user endpoint

karenmillermoj:


helm_deploy/hmpps-auth/Chart.yaml
```diff
@@ -6,8 +6,8 @@ version: 0.1.1
 
 dependencies:
   - name: generic-service
-    version: 2.1.0
+    version: 2.2.0
```
SimonMitchellMOJ, line 9: updating to 2.2.0 means we can remove the    
    ```
v1_2_enabled: false
    v0_47_enabled: true
```
from all the helm values files

karenmillermoj: oo - will do that FANTASTIC!!!

ravmoj (approved):
lgtm

PR #1449, opened 2023-01-17
HAAR-862 token expiry extension and reset

AlTurner-MOJ:


src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/security/UserService.kt
```diff
@@ -228,6 +232,9 @@ class UserService(
   }
 }
 
+class UserNotFoundException(val userId: UUID) :
+  Exception("User: ${userId.toString().uppercase()} not found")
```
karenmillermoj, line 236: Should this be a RuntimeException, like the UsernameNotFoundException

src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/resource/api/TokenController.kt
```diff
@@ -98,6 +100,42 @@ class TokenController(private val tokenService: TokenService) {
     @Valid @RequestBody
     tokenByEmailTypeRequest: TokenByEmailTypeRequest
   ) = tokenService.createTokenByEmailType(tokenByEmailTypeRequest)
+
+  @PreAuthorize("hasRole('ROLE_CREATE_EMAIL_TOKEN')")
+  @PostMapping("/api/token/reset/{userId}")
+  @Operation(
```
karenmillermoj, line 106: Should this just be a PUT?  We are updating and there is no body

AlTurner-MOJ: Possibly, though the other end points in there are all POST. I think it creates a new UserToken record?

src/test/kotlin/uk/gov/justice/digital/hmpps/oauth2server/verify/TokenServiceTest.kt
```diff
@@ -281,4 +286,28 @@ class TokenServiceTest {
...
+      whenever(userService.getUserById(userId)).thenReturn(user)
+
+      val token = tokenService.createResetTokenForUserWithSevenDayExpiry(userId)
+
+      assertNotNull(token)
+    }
+
+    @Test
+    fun `Create token user not found`() {
+      doThrow(UsernameNotFoundException("failed")).whenever(userService).getUserById(userId)
```
karenmillermoj, line 306: Isn't this a UserNotFoundException exception that is thrown

src/test/kotlin/uk/gov/justice/digital/hmpps/oauth2server/resource/api/TokenControllerIntTest.kt
```diff
@@ -188,4 +188,55 @@ class TokenControllerIntTest : IntegrationTest() {
...
+
+      webTestClient
+        .post().uri("/api/token/reset/C9999EE9-99BF-999F-999C-AA99C9DF99F9")
+        .headers(setAuthorisation("AUTH_ADM", listOf("ROLE_CREATE_EMAIL_TOKEN")))
+        .exchange()
+        .expectStatus().isNotFound
+        .expectBody()
+        .json(
+          """
+        {"error":"Not Found","error_description":"User: C9999EE9-99BF-999F-999C-AA99C9DF99F9 not found","field":"userId"}
```
karenmillermoj, line 222: prob separate the json to make more readable

src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/resource/api/TokenController.kt
karenmillermoj: If this is a create and not just an update, then that's fine.

src/test/kotlin/uk/gov/justice/digital/hmpps/oauth2server/resource/api/TokenControllerIntTest.kt
AlTurner-MOJ: Re-formatted - does that help?

src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/security/UserService.kt
```diff
@@ -117,8 +118,9 @@ class UserService(
   fun getUser(username: String): User =
     findUser(username).orElseThrow { UsernameNotFoundException("User with username $username not found") }
 
+  @Throws(UserNotFoundException::class)
```
karenmillermoj, line 121: Do we need the @throws? and does it need   @Transactional(readOnly = true)

AlTurner-MOJ: The transactional boundary for this unit of work is the calling class above it, TokenService. This read operation will be executed as part of that transactional scope, which is not read only (the User entity is changed by addition of the new token).
As I understand it you do need the @Throws annotation for checked exceptions? I've left this as a checked exception to align with other similar exception types in Auth i.e. RoleNotFoundException

AlTurner-MOJ: Removed @Throws and added @Transactional hint

karenmillermoj (approved):
lgtm

PR #1458, opened 2023-02-02
HAAR-1167: ✨ add deployment information free text field

SimonMitchellMOJ:


src/main/resources/db/auth/V1_6__client_details_extra_info.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE OAUTH_CLIENT_DEPLOYMENT_DETAILS ADD deployment_info varchar(1000) null;
```
karenmillermoj, line 1: By default the column is nullable so you wouldn't need the "null" on the end.  It looks like it doesn't affect it though.

src/main/resources/templates/ui/deploymentForm.html
```diff
@@ -126,9 +126,16 @@ <h1 class="govuk-heading-l">Client deployment information</h1>
...
         </fieldset>
+
       </div>
+            <div class="govuk-form-group">
+              <label class="govuk-label" for="deploymentInfo">Deployment info</label>
+              <textarea class="govuk-textarea" id="deploymentInfo" name="deploymentInfo" type="text"
+                        rows="6"
+                        th:text="${clientDeployment.deploymentInfo}">info</textarea>
+            </div>
+
```
karenmillermoj, line 138: formatting looks a bit odd here?

karenmillermoj (dismissed):
lgtm

src/main/resources/templates/ui/deploymentForm.html
```diff
@@ -129,13 +129,12 @@ <h1 class="govuk-heading-l">Client deployment information</h1>
...
-              <textarea class="govuk-textarea" id="deploymentInfo" name="deploymentInfo" type="text"
-                        rows="6"
-                        th:text="${clientDeployment.deploymentInfo}">info</textarea>
-            </div>
-
+        <div class="govuk-form-group">
+          <label class="govuk-label" for="deploymentInfo">Deployment info</label>
+          <textarea class="govuk-textarea" id="deploymentInfo" name="deploymentInfo" type="text"
+                    rows="6"
+                    th:text="${clientDeployment.deploymentInfo}">info</textarea>
```
karenmillermoj, line 136: Still a bit off but not as bad - the rows="6" could now go on the previous line - but minor improvement

PR #1459, opened 2023-02-03
HAAR-884 Provide endpoint to find by username and source - which also saves to db if doesn't already exist

karenmillermoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/resource/api/UserController.kt
```diff
@@ -141,6 +141,44 @@ class UserController(private val userService: UserService) {
     val backupVerified: Boolean
   )
 
+  @GetMapping("/api/user")
+  @Operation(
+    summary = "Finds a user based on their username and source and returns their User detail. " +
```
karenmillermoj, line 146: Not convinced on the endpoint name - any suggestions welcome

AlTurner-MOJ: Struggling to think of alternative

AlTurner-MOJ (approved):
Couple of comments but looks good

src/test/kotlin/uk/gov/justice/digital/hmpps/oauth2server/resource/api/UserControllerTest.kt
```diff
@@ -97,6 +104,96 @@ class UserControllerTest {
     )
   }
 
+  @Test
```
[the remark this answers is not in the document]
karenmillermoj: you're probably correct.  Possibly 2 different successes - one for Nomis and one for non-nomis (as need to verify additional data).  Will remove Azure and Delius

AlTurner-MOJ (approved):
lgtm

PR #1461, opened 2023-02-06
HAAR-868: Deprecate '/api/authuser/me/searchable-roles'

ravmoj:
Deprecate '/api/authuser/me/searchable-roles'


karenmillermoj (approved):
lgtm

PR #1462, opened 2023-02-07
HAAR-1167: add deployment info to client view only page

SimonMitchellMOJ:


src/main/resources/templates/ui/viewOnlyForm.html
```diff
@@ -348,6 +348,15 @@ <h2 class="govuk-heading-m">Deployment details</h2>
         Secret key
       </dd>
     </div>
+    <div class="govuk-summary-list__row" th:if="${#strings.equals(deployment.hosting, 'CLOUDPLATFORM')}">
```
karenmillermoj, line 351: It should be shown for non-cloud platform too

karenmillermoj (dismissed):
lgtm

PR #1468, opened 2023-02-16
HAAR-1149: 🔨 update client rotation script to have variable if limited…

SimonMitchellMOJ:
… special characters need in client secret


karenmillermoj (approved):
lgtm

PR #1475, opened 2023-03-02
HAAR-881 Fetching Azure User only returns azure user specific data

karenmillermoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/resource/api/AzureUserController.kt
```diff
@@ -63,10 +62,26 @@ class AzureUserController(
...
-    val user = userService.getOrCreateUser(username.toString()).orElseThrow()
-    return UserDetail.fromPerson(upd, user)
-  }
+  ) = azureUserService.getAzureUserByUsername(username.toString().uppercase()).map {
+    AzureUser(
+      username = UUID.fromString(it.username),
+      email = it.email,
+      enabled = it.isEnabled,
+      firstName = WordUtils.capitalizeFully(it.firstName),
+      lastName = WordUtils.capitalizeFully(it.surname),
```
karenmillermoj, line 71: We are now just returning data specific to the Azure User.  This is currently only called my manage-users-api.

AlTurner-MOJ (approved):
lgtm

PR #1480, opened 2023-03-06
HAAR-1244: upgrade spring boot plugin 4.8.4 beta 2

SimonMitchellMOJ:


karenmillermoj (approved):
lgtm

src/test/kotlin/uk/gov/justice/digital/hmpps/oauth2server/resource/api/RolesControllerIntTest.kt
```diff
@@ -737,6 +738,14 @@ class RolesControllerIntTest : IntegrationTest() {
         .body(fromValue(mapOf("roleName" to "new role name")))
         .exchange()
         .expectStatus().isOk
+
+      // reset role
+      webTestClient
```
karenmillermoj, line 743: Could have this as a function to call - but works fine as is

PR #1486, opened 2023-03-16
HAAR-1267: new end point supporting prison user email address update

AlTurner-MOJ:


src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/resource/api/PrisonUserController.kt
```diff
@@ -177,6 +178,14 @@ class PrisonUserController(
     return if (smokeTestEnabled) userEmailAndUsername.link else ""
   }
 
+  @PutMapping("/{username}/email")
+  @PreAuthorize("hasAnyRole('ROLE_MAINTAIN_ACCESS_ROLES_ADMIN')")
```
karenmillermoj, line 182: Does this need any swagger docs adding

AlTurner-MOJ: Yes, sorry! Added

src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/security/NomisUserService.kt
```diff
@@ -37,6 +38,13 @@ class NomisUserService(
     return allNomis.map { it.copy(email = emailLowered) }
   }
 
+  @Transactional
```
karenmillermoj, line 41: Is @transactional needed - as I think crudrespository methods are transactional by default?

karenmillermoj (approved):
lgtm - minor comment - annotation can probably be removed (but not essential)

src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/security/NomisUserService.kt
AlTurner-MOJ: Not 100% but think it is needed here - note I'm not calling save on the repository. I believe Hibernate flushes all updated entities on transaction commit

PR #1487, opened 2023-03-17
HAAR-1275: removing email domain code

olatomoloju:
This relates to removing email domain code.
NB:

1. The email domain service tile will be remove via endpoint in relevant environment, rather than deleting record from oauth_service db table.
2. The repo class is not removed, because it has dependency with the email domain cache and email verification service, which is in turn used by multiple endpoints.


SimonMitchellMOJ (approved):
LGTM

PR #1497, opened 2023-03-27
HAAR-1276: Adding client id and client secret

ravmoj:
Adding client id and client secret to invoke manage-users-api-service from hmpps-auth


karenmillermoj (approved):
lgtm

PR #1505, opened 2023-03-31
HAAR-1308 Sync Nomis email back to auth not called via manage-users-a…

karenmillermoj:
…pi, passing in nomis email


ravmoj (approved):
lgtm

PR #1506, opened 2023-04-06
HAAR-864: Alter tables to autogenerate primary key 

ravmoj:
Alter tables to autogenerate primary key by default for User table


src/main/resources/db/auth_postgresql/V1_4__auto_create_id.sql
```diff
@@ -4,3 +4,4 @@ ALTER TABLE groups ALTER group_id SET DEFAULT gen_random_uuid();
 ALTER TABLE roles ALTER role_id SET DEFAULT gen_random_uuid();
 ALTER TABLE email_domain ALTER email_domain_id SET DEFAULT gen_random_uuid();
 ALTER TABLE child_group ALTER child_group_id SET DEFAULT gen_random_uuid();
+ALTER TABLE users ALTER user_id SET DEFAULT gen_random_uuid();
```
karenmillermoj, line 7: Because this is file is used in the main deployment (not just dev data) the change will need to be in a separate flyway script

karenmillermoj (approved):
lgtm

PR #1507, opened 2023-04-17
HAAR-1343 update external user email

AlTurner-MOJ:


src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/resource/api/AuthUserSyncController.kt
```diff
@@ -0,0 +1,52 @@
...
+ */
+@RestController
+class AuthUserSyncController(
+  val authUserSyncService: AuthUserSyncService,
+  val telemetryClient: TelemetryClient,
+) {
+
+  @PutMapping("/api/authuser/sync/{username}/email")
+  @PreAuthorize("hasAnyRole('ROLE_MAINTAIN_OAUTH_USERS', 'ROLE_AUTH_GROUP_MANAGER')")
+  fun emailUpdate(@PathVariable username: String, @RequestBody emailUpdate: UserEmailUpdate) {
```
ravmoj, line 28: Add swagger docs

src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/maintain/AuthUserSyncService.kt
```diff
@@ -0,0 +1,20 @@
...
+@Service
+@Transactional
+class AuthUserSyncService(
+  private val userRepository: UserRepository,
+) {
+  fun syncUserEmail(username: String, emailUpdate: UserEmailUpdate) {
+    val user = userRepository.findByUsername(username).orElseThrow { UsernameNotFoundException("User with name $username not found") }
+    user.email = emailUpdate.email
+    user.username = emailUpdate.username
+    user.verified = false
```
ravmoj, line 18: why is it false?

src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/resource/api/AuthUserSyncController.kt
```diff
@@ -0,0 +1,52 @@
...
+ * Although part of the Auth API, these end points are intended to be used exclusively from hmpps-manage-users-api
+ * to keep external user data in sync, and therefore are not part of the published API.
+ */
+@RestController
+class AuthUserSyncController(
+  val authUserSyncService: AuthUserSyncService,
+  val telemetryClient: TelemetryClient,
+) {
+
+  @PutMapping("/api/authuser/sync/{username}/email")
```
ravmoj, line 26: do we need to add role for endpoints to access internal services?

src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/resource/api/AuthUserSyncController.kt
AlTurner-MOJ: This was discussed with @karenmillermoj - the intention of this (and other) sync end points is that they will only be used from manage-users-api for the purpose of keeping user related data sychronised between the Auth and external-users databases. These end points will not form part of the 'published' Auth API, which is why there is no Swagger documentation.
The roles on this end point are the same as those required to use the single client context (amend a user email)

src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/maintain/AuthUserSyncService.kt
AlTurner-MOJ: This is a little bit of business logic which you are right to question. I wanted to make this end point essentially dumb in the sense that it just does what the client asks (and assumes that the client has completed any necessary validations/checks). I debated whether to allow the client to pass the verified flag in, but the only use case for this is altering a user's email, and we always set the verified flag to false when this happens.

src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/resource/api/AuthUserSyncController.kt
AlTurner-MOJ: Possibly - I left the role the same as that required to access the single intended client use case (amend user email). Did not feel like adding a specific separate role for this should be necessary?

ravmoj (approved):
lgtm

PR #1511, opened 2023-04-20
HAAR-1364 Upgrade db to postgres 15

karenmillermoj:


ankit-gautam23 (commented):
Version bumping is good. But plz check once with the API contracts it shouldn't break because of this

karenmillermoj:
Duplicate

PR #1514, opened 2023-04-25
PI-987 Update local sentence-plan clients

marcus-bcl:
Updates the secret to match the id, and adds localhost:3000/sign-in/callback as a registered redirect URI.


SimonMitchellMOJ (approved):
LGTM

PR #1521, opened 2023-05-18
HAAR-1452 Call to get userId doesn't need to call nomis if already in auth database

karenmillermoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/security/UserService.kt
```diff
@@ -151,6 +151,21 @@ class UserService(
       }
     }
 
+  @Transactional
+  fun getOrCreateUser(username: String, source: AuthSource): Optional<User> =
+    findUser(username, source).or {
+      getMasterUserPersonDetails(username, source).map {
+        val user = it.toUser()
+        if (AuthSource.fromNullableString(user.authSource) == nomis) {
+          getEmailAddressFromNomis(username).ifPresent { email ->
```
ravmoj, line 160: getEmailAddressFromNomis is common method  for other services and does additional call to "nomisUserService.getNomisUserByUsername(username)" can we move this nomis specific logic here 
 if (isHmpsGsiEmail(it)) Optional.empty() else Optional.of(it)

karenmillermoj: I totally agree - but this is done in so many places.  Are you happy if I raise this as a separate ticket?
@ravmoj

ravmoj: sure @karenmillermoj

ravmoj (approved):
lgtm

PR #1526, opened 2023-05-25
Revert "HAAR-1454: allow azure user to access service with no roles (…

SimonMitchellMOJ:
…#1525)"

This reverts commit ac7568da3f9ac349c31ca5842c0c60a742453d2f.

PR #1527, opened 2023-05-25
HAAR-1454: allow azure user to access service with no roles 

SimonMitchellMOJ:
Reverts ministryofjustice/hmpps-auth#1526


src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/service/UserContextService.kt
```diff
@@ -35,9 +35,16 @@ class UserContextService(
     val requestedSources = findRequestedSources(scopes)
 
     val email = userService.getEmail(loginUser) ?: return emptyList()
-    return requestedSources.map { findEnabledUsersWithRoles(email, it, roles) }
+
+    val userList = requestedSources.map { findEnabledUsersWithRoles(email, it, roles) }
       .filter { it.isNotEmpty() }
       .flatten()
+    // if it doesn't find any users and roles is empty then the azure user
```
karenmillermoj, line 42: not sure what the comment means -> **then the azure user** does it mean "then is an azure user"

karenmillermoj (approved):
lgtm

src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/service/UserContextService.kt
SimonMitchellMOJ: updated

PR #1529, opened 2023-05-30
HAAR-1403: Remove user tokens of type MFA_CODE and MOBILE 

ravmoj:
Remove user tokens of type MFA_CODE and MOBILE older than one day of token expiry


SimonMitchellMOJ (approved):
LGTM

PR #1544, opened 2023-07-18
HAAR-1499: Remove all expired tokens after they have expired

ravmoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/auth/repository/UserTokenRepository.kt
```diff
@@ -15,6 +15,6 @@ interface UserTokenRepository : CrudRepository<UserToken, String>, JpaSpecificat
 
   @Modifying
   @Enumerated
-  @Query("DELETE FROM UserToken ut where ut.tokenType in( 'MFA_CODE', 'MOBILE' ) and ut.tokenExpiry < :tokenExpiry")
+  @Query("DELETE FROM UserToken ut where ut.tokenExpiry < :tokenExpiry")
```
karenmillermoj, line 18: You could prob rename this method to do what you want (i.e. without the need to have the @Query

karenmillermoj (approved):
lgtm

PR #1548, opened 2023-07-24
HAAR-1223: Moving JWKS ids and aliases to secrets for ease of rotation

brightonsbox:
I want to be able to quickly rotate primary and auxiliary keys round using kubernetes secrets rather than keeping them in values.yml files.  I know I can edit the deployment through k8s but it will be a lot less of a faff this way, and everything is kept together in one place.


ravmoj (approved):
lgtm

PR #1552, opened 2023-07-31
HAAR-842: To suppress alerts related to  'session attribute authorizationRequest'

ravmoj:
To suppress alerts related to  'session attribute authorizationRequest' (part of HttpSessionRequiredException)


brightonsbox (approved):
LGTM

PR #1555, opened 2023-08-09
HAAR-1682: Reverted DB refresh schedule

ravmoj:
Reverted schedule back to 10:30 on every Sunday on every even week.


karenmillermoj (approved):
lgtm

brightonsbox (approved):
LGTM

PR #1567, opened 2023-09-12
HAAR-1805: ➖ supress veracode dependency vulnerability and update bouncycastle dependency

SimonMitchellMOJ:


src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/resource/ResetPasswordController.kt
```diff
@@ -94,7 +93,7 @@ class ResetPasswordController(
           modelAndView.addObject("resetLink", resetLink.get())
         }
       } else {
-        log.info("Reset password request failed, no link provided for $usernameOrEmail".removeAllCrLf())
+        log.info("Reset password request failed, no link provided for $usernameOrEmail")
```
karenmillermoj, line 96: hmm odd that the StringSanitiser is causing an issue.  It was added because of a veracode vulnerability.  Is it worth speaking to the veracode guy (can't remember his name)

ravmoj (approved):
LGTM

PR #1570, opened 2023-09-26
HAAR-1871: add product id to info endpoint

SimonMitchellMOJ:


src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/service/ProductIdInfoContributor.kt
```diff
@@ -0,0 +1,14 @@
+package uk.gov.justice.digital.hmpps.oauth2server.service
+
+import org.springframework.beans.factory.annotation.Value
+import org.springframework.boot.actuate.info.Info
+import org.springframework.boot.actuate.info.InfoContributor
+import org.springframework.stereotype.Component
+
+@Component
+class ProductIdInfoContributor(@Value("\${product-Id:default}") private val productId: String) : InfoContributor {
```
petergphillips, line 9: Would prefer `product-id` here instead.  Also should this file be in the `config` rather than `service` directory?

petergphillips (dismissed):
LGTM

petergphillips (approved):
LGTM

PR #1571, opened 2023-09-28
[HAAR-1878] Add Accredited Programmes authorization_code client

Gweaton:
It looks like we've somehow been using the `approved-premises-ui` client locally without realising. This should allow us to use our own.

https://dsdmoj.atlassian.net/browse/HAAR-1878


ravmoj (approved):
lgtm

PR #1576, opened 2023-10-04
PI-1493 Switch from Community API to Probation Integration service

marcus-bcl:
The client currently used to talk to Community API will need the new role of `ROLE_DELIUS_USER_AUTH`.  Once the changes are deployed you can then remove `ROLE_COMMUNITY_AUTH_INT`.


SimonMitchellMOJ (approved):
LGTM

PR #1590, opened 2023-11-01
HAAR-1890: NOMIS user sign in updates the LAST_LOGON_DATE in NOMIS

brightonsbox:
The method` resetRetriesAndRecordLogin` is called either when logging into HMPPS Auth with NOMIS credentials or after having logged in with DOM1 account and selected a NOMIS user to use for access to a service through the authorization_code flow.

By adding an API call to `nomis-user-roles-api` to update the `last_logon_date` here, this solves the problem with DOM1 users getting made inactive. If they use HMPPS Digital services via HMPPS Auth, the `last_logon_date` will be updated which is used by the update inactivity batch job to determine inactivity.


src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/security/UserRetriesService.kt
```diff
@@ -56,6 +56,8 @@ class UserRetriesService(
       is NomisUserPersonDetails -> {
         // Copy first name and surname for Nomis users (don't overwrite email address)
         user.person = Person(userPersonDetails.firstName, userPersonDetails.surname)
+        // Update the LAST_LOGON_DATE in the STAFF_USER_ACCOUNTS table in NOMIS - user inactivity job runs using this
+        userService.recordNomisSignIn(username)
```
brightonsbox, line 60: This is the important line here.

The method `resetRetriesAndRecordLogin` is called either when logging into HMPPS Auth with NOMIS credentials or after having logged in with DOM1 account and selected a NOMIS user to use for access to a service through the `authorization_code` flow.

This solves the problem with DOM1 users getting made inactive.  If they use HMPPS Digital services via HMPPS Auth, the `last_logon_date` will be updated which is used by the update inactivity batch job to determine inactivity.

src/test/kotlin/uk/gov/justice/digital/hmpps/oauth2server/security/UserRetriesServiceTest.kt
```diff
@@ -31,149 +36,126 @@ class UserRetriesServiceTest {
 
   @Nested
   inner class resetRetriesAndRecordLogin {
-    @Test
-    fun resetRetriesAndRecordLogin() {
-      whenever(userService.getEmailAddressFromNomis(anyString())).thenReturn(Optional.of("bob@bob.justice.gov.uk"))
-      service.resetRetriesAndRecordLogin(userPersonDetailsForBob)
-      verify(userRetriesRepository).save<UserRetries>(
+    @ParameterizedTest
```
brightonsbox, line 39: The tests here were a little poor and some of these needed to be tested against all UserPersonDetails types, so I've parameterized the tests

src/test/kotlin/uk/gov/justice/digital/hmpps/oauth2server/security/NomisAuthenticationProviderTest.kt
```diff
@@ -82,7 +82,7 @@ class NomisAuthenticationProviderTest {
   }
 
   @Test
-  fun `authenticate ResetAfterSuccess`() {
+  fun `authenticate_ResetAfterSuccess`() {
```
SimonMitchellMOJ, line 85: Why the _ the method name is in ``?

SimonMitchellMOJ (approved):
LGTM

src/test/kotlin/uk/gov/justice/digital/hmpps/oauth2server/security/NomisAuthenticationProviderTest.kt
brightonsbox: Yeah... no reason, must have started fiddling and then decided against it.

PR #1609, opened 2023-11-27
Removing call to update client last_accessed as a temporary emergency fix

ravmoj:


ravmoj:
No longer required , as could able to merge PR 1608

PR #1617, opened 2023-12-11
Introducting prompt for justice.gov.uk users to change NOMIS password

brightonsbox:


nwilliamsmoj (commented):
What happens after they have chnaged their password - I can see that it redirects to 
return ModelAndView("forward:/oauth/authorize").

Will the redirect just go through ACF again without any need to Login as they have just done that?

brightonsbox:
> What happens after they have chnaged their password - I can see that it redirects to return ModelAndView("forward:/oauth/authorize").
> 
> Will the redirect just go through ACF again without any need to Login as they have just done that?

Yes, once they change their password, the authorize flow retries basically, but now because they've reset the password, it won't force them to do it again.

SimonMitchellMOJ (approved):
LGTM

